PR #79, opened 2019-01-04
document viewer tests to enzyme

B3zo0:


packages/sn-document-viewer-react/test/DocumentTitlePager.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <DocumentTitlePagerComponent
+        activePages={[1]}
+        documentName="TestDocument"
+        gotoPage="goto"
+        pageCount={10}
+        setActivePages={jest.fn()}
+      />,
+    )
+    wrapper.find(Typography).simulate('click')
+    await sleepAsync()
```
gallayl: I think the sleep is not neccessary but we should rather make an assertion if a child element exists.
```suggestion
    expect(wrapper.find(TextField)).toBeTruthy()
```

packages/sn-document-viewer-react/test/DocumentTitlePager.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <DocumentTitlePagerComponent
+        activePages={[1]}
+        documentName="TestDocument"
+        gotoPage="goto"
+        pageCount={10}
+        setActivePages={activePages}
+      />,
+    )
+    wrapper.find(Typography).simulate('click')
+    await sleepAsync()
```
gallayl: We should remove this sleep too...

PR #80, opened 2019-01-08
Feature/update dependencies

gallayl:
- [x] Update and consolidate all dependencies
 - [x] Fix deprecation warnings
 - [x] Fix missing source mapping warning

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,7 @@
     "packages": [
       "packages/*",
       "examples/*"
-    ],
-    "nohoist": []
+    ]
   },
   "config": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -6,12 +6,7 @@
       "packages/*",
       "examples/*"
     ],
-    "nohoist": [
-      "**@types/**",
-      "**/react-moment",
-      "**/material-ui-pickers",
-      "**/react-markdown"
-    ]
+    "nohoist": []
```
B3zo0: I think we can remove this from here.

PR #95, opened 2019-01-29
Search/Reference field refactor

gallayl:
[search-react]
 - added react-autosuggest library for reference fields

[document-viewer-react]
 - added snapshot tests for actions

### DIFF
--- a/packages/sn-search-react/package.json
+++ b/packages/sn-search-react/package.json
@@ -38,7 +38,7 @@
     "@types/react-dom": "^16.0.11",
     "@types/react-router-dom": "^4.3.1",
     "@types/uuid": "^3.4.4",
-    "@types/autosuggest-highlight": "^3.1.0",
+    "@types/autosuggest-highlight": "3.1.0",
     "@types/react-autosuggest": "^9.3.6",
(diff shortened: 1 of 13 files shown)


packages/sn-search-react/package.json
```diff
@@ -38,6 +38,8 @@
     "@types/react-dom": "^16.0.11",
     "@types/react-router-dom": "^4.3.1",
     "@types/uuid": "^3.4.4",
```
B3zo0: we don't use uuid in this package
```suggestion
```

packages/sn-search-react/src/Components/Fields/ReferenceField.tsx
```diff
@@ -1,78 +1,93 @@
...
 
 /**
  * Reference field picker component
  */
 export class ReferenceField<T extends GenericContent = GenericContent> extends React.Component<
-  ReferenceFieldProps<T> & MaterialTextFieldProps,
+  ReferenceFieldProps<T>,
   ReferenceFieldState<T>
 > {
+  /**
```
B3zo0: Please remove the empty comment.

PR #103, opened 2019-02-04
Feature/pickers react

B3zo0:
Pickers react package with list picker component


packages/sn-pickers-react/src/ListPicker/ListPicker.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { useAsync } from 'react-use'
+import { ItemComponent } from './Item'
+
+/**
+ * Properties for list picker component.
+ * @interface ListPickerProps
+ * @template T
+ */
+export interface ListPickerProps<T extends GenericContent = GenericContent> {
+  loadItems: (path: string) => Promise<T[]>
```
herflis: Please add comments for all the props

PR #123, opened 2019-03-08
Avatar control

herflis:


packages/sn-controls-react/src/fieldcontrols/Avatar/DefaultAvatarTemplate.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }
+  public handleAddIconClick = () => {
+    this.props.add()
+  }
+  /**
+   * render
+   * @return {ReactElement} markup
+   */
+  public render() {
+    const { url } = this.props
```
B3zo0: Would it make sense to destruct all the props that are used in the render method?

PR #126, opened 2019-03-12
[KFI]fix(dms-demo): Fix workspace selector favorites

herflis:
Favorite ws list will be empty after user logout


kultsar:
@herflis, seems good for me and would like to approve, but I'm confused by "Favorite ws list will be empty after user logout". Does it refer to a part of the implementation?

herflis:
@kultsar It was the solution to make this list empty. The former problem was, that after a user logged out, her favorites were not cleared in the browser/on the frontend and than if a new user logged-in, the favoritelist contained not only the new users favorites but the logged-out user's too.

PR #129, opened 2019-03-14
Groups page

herflis:


examples/sn-dms-demo/src/components/UsersAndGroups/UserSelector/UserSearch.tsx
```diff
@@ -0,0 +1,110 @@
...
+  closeDropDown: (open: boolean) => void
+}
+
+class UserSearch extends React.Component<UserSearchProps & WithStyles<C>, {}> {
+  constructor(props: UserSearch['props']) {
+    super(props)
+    this.handleKeyup = this.handleKeyup.bind(this)
+  }
+  public handleKeyup = (e: React.ChangeEvent<HTMLInputElement>) => {
+    // tslint:disable-next-line:no-string-literal
```
B3zo0:
```suggestion
```

examples/sn-dms-demo/src/components/UsersAndGroups/UserSelector/UserListItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+interface UserListItemState {
+  selected: boolean
+}
+
+class UserListItem extends React.Component<
+  { classes: any } & ReturnType<typeof mapStateToProps> & typeof mapDispatchToProps & UserListItemProps,
+  UserListItemState
+> {
+  public state = {
+    selected: this.props.selected,
```
B3zo0: You should use the props.selected value instead of using it as a state because this way it is not working as expected. It is not going to be updated when the props change.
The simplest solution is to use the props.selected state. The other one is to use getderivedstatefromprops.

examples/sn-dms-demo/src/pages/Dashboard.tsx
```diff
@@ -201,9 +201,23 @@ class DashboardComponent extends React.Component<
                             </Switch>
                           )}
                         />
-                        <Route path="/groups">
-                          <Groups />
-                        </Route>
+                        <Route
+                          path="/groups"
+                          // tslint:disable-next-line: no-unnecessary-type-annotation
```
B3zo0:
```suggestion
```

PR #152, opened 2019-04-02
[KFI]fix(pickers-react): list picker items prop

### LINKED ISSUE #125: [FEATURE] ListPicker reload/refresh
# 🙋 Feature Request

It will be great if the listpicker's itemlist could be reloaded/refreshed from the parent component (e.g. after a fileupload, filtering, etc).

## 📦 Package / Application

@sensenet/pickers-react

## 📃 Version

1.0.0

## 🤔 Expected Behavior

The inner children list of the picker should be reloaded/refreshed somehow from the parent component.

## 😯 Current Behavior

The list in the picker remains the same after calling loadItems or loadParent from the parent component.

## 💁 Possible Solution

It's ok if there's a method that can be called to reload the children list or if we can set a static itemlist as children.

## 🔦 Context

I'm currently working on an avatarpicker, where the user can upload a new image on the picker and when the upload was succesful she has to be able to pick the newly uploaded image too.

B3zo0:
## What I did in this pr
- refactored upload and made it available in the repository object **This is a breaking change** 💥
- added new repository events `onUpload` ~this needs attention~
- moved pickers react example app to storybook
- updated @types/styled-components to version 4.1.13 because of (https://github.com/Microsoft/TypeScript/issues/30663)
- reworked list picker component with the tasks below

## Tasks
- [x] Use repository events to reload items
- [x] Add odata otpions property for loadItems and parent (combineOdataFieldParams)
- [x] Handle errors inside component
- [x] Remove items, loadItems, loadParent props use oData options instead

Fixes #125


packages/sn-repository-events/src/EventHub.ts
```diff
@@ -120,6 +120,7 @@ export class EventHub implements Disposable {
    * Triggered after copying a content has been failed
    */
   public readonly onContentCopyFailed = new ObservableValue<ContentCopyFailed>()
+  public readonly onUpload = new ObservableValue<{ isUploaded: boolean }>()
```
gallayl:
```suggestion
  public readonly onUploadFinished = new ObservableValue<{ isUploaded: boolean }>()
```
I think an onUploadFailed can be also useful.

B3zo0: 👍

packages/sn-client-core/src/Repository/Upload.ts
```diff
@@ -46,11 +46,12 @@ export interface UploadResponse {
  * Helper class for uploading content into the sensenet Repository
  */
 export class Upload {
+  constructor(private repository: Repository) {}
```
gallayl:
```suggestion
  constructor(private readonly repository: Repository) {}
```

packages/sn-client-core/src/Repository/Upload.ts
```diff
@@ -74,20 +75,20 @@ export class Upload {
    * @param {File} file The File object
    * @param {Repository} repo The sensenet Repository
    */
-  public static isChunkedUploadNeeded(file: File, repo: Repository): boolean {
+  public isChunkedUploadNeeded(file: File, repo: Repository): boolean {
```
gallayl: I think we can clean up the API from this unneccessary parameter
```suggestion
  public isChunkedUploadNeeded(file: File): boolean {
```

packages/sn-client-core/src/Repository/Upload.ts
```diff
@@ -74,20 +75,20 @@ export class Upload {
...
     return file.size >= repo.configuration.chunkSize
```
gallayl:
```suggestion
    return file.size >= this.repo.configuration.chunkSize
```

packages/sn-client-core/src/Repository/Upload.ts
```diff
@@ -62,8 +63,8 @@ export class Upload {
    * Uploads a specified file into a sensenet Repository
    * @param {UploadFileOptions} options The additional upload options
    */
-  public static async file<T extends Content>(options: UploadFileOptions<T>): Promise<UploadResponse> {
-    if (this.isChunkedUploadNeeded(options.file, options.repository)) {
+  public async file<T extends Content>(options: UploadFileOptions<T>): Promise<UploadResponse> {
+    if (this.isChunkedUploadNeeded(options.file, this.repository)) {
```
gallayl:
```suggestion
    if (this.isChunkedUploadNeeded(options.file)) {
```

packages/sn-pickers-react/src/ListPicker/ListPicker.tsx
```diff
@@ -18,21 +19,40 @@ export function ListPickerComponent<T extends GenericContent = GenericContent>(p
...
 
-  const loadItems = () =>
-    (props.loadItems && props.loadItems(currentPath)) || defaultLoadItems(currentPath, props.repository)
-
-  const loadParent = () =>
-    (props.loadParent && props.loadParent(parentId)) || defaultLoadParent(parentId, props.repository)
-
-  const { loading, value: items, error } = useAsync(loadItems, [currentPath])
-  const parent = useAsync(loadParent, [parentId])
+  const update = () => reload()
```
gallayl: There can be many events that triggers unneccessary reloads (e.g. uploading many small files). Can we debounce reload to improve performance?

B3zo0: I've added debounce there.

gallayl:
We should also notice that this will be a breaking change in the next release of *client-core*.

PR #161, opened 2019-04-05
Admin UI Content editing improvements

gallayl:


apps/sensenet/src/components/MainRouter.tsx
```diff
@@ -30,35 +29,31 @@ const MainRouter: React.StatelessComponent<RouteComponentProps> = () => {
...
-                path="/:repo/editBinary/:contentId?"
-                render={() => <EditBinary />}
-                authorize={() => true}
-              />
-              <AuthorizedRoute
+              <Route path="/:repo/search" render={() => <SearchComponent />} />
+              <Route path="/:repo/iam" render={() => <IamComponent />} />
+              <Route path="/:repo/setup" render={() => <SetupComponent />} />
+              <Route path="/:repo/info" render={() => <VersionInfoComponent />} />
+              <Route path="/:repo/editBinary/:contentId?" render={() => <EditBinary />} authorize={() => true} />
```
B3zo0: Is the authorize prop still needed here? Same q for line 54, 59

gallayl: You're right, I've removed them

PR #171, opened 2019-04-18
Fix/tsc build

B3zo0:
## What I did in this pr:
- build improvement
  - build all packages with project references in packages folder (yarn build now runs `tsc -b packages`)
  - moved icons assets outside of src so it is not needed to be copied to dist
- controls-react
  - comment styling and typo fixes
  - removed unused files
  - removed unused dependecies


packages/sn-controls-react/package.json
```diff
@@ -62,19 +59,11 @@
...
-    "object-assign": "4.1.1",
-    "postcss-flexbugs-fixes": "4.1.0",
-    "promise": "^8.0.3",
     "prop-types": "^15.7.2",
     "react": "^16.8.4",
-    "react-dev-utils": "^8.0.0",
-    "react-dom": "^16.8.4"
+    "react-dom": "^16.8.4",
+    "react-redux": "^6.0.1",
+    "redux": "^4.0.1"
```
gallayl: I think we should refactor the *controls-react* package to work without redux in the future

PR #177, opened 2019-04-24
Remove redux and related from the react-controls package

herflis:


packages/sn-controls-react/src/viewcontrols/NewView.tsx
```diff
@@ -173,8 +160,4 @@ export class NewViewComponent<T extends GenericContent> extends Component<
   }
 }
 
-const newView = connect(
-  mapStateToProps,
-  mapDispatchToProps,
-)(NewViewComponent)
-export { newView as NewView }
+export { NewViewComponent as NewView }
```
B3zo0: I think it would make sense to rename the component or remove this line. I don't see a value exporting this in 2 different name.

packages/sn-controls-react/src/viewcontrols/NewView.tsx
```diff
@@ -3,28 +3,24 @@
...
 import Grid from '@material-ui/core/Grid'
 import Typography from '@material-ui/core/Typography'
 import { Repository } from '@sensenet/client-core'
 import { ControlSchema } from '@sensenet/control-mapper'
 import { GenericContent, Schema } from '@sensenet/default-content-types'
-import { Actions, Reducers } from '@sensenet/redux'
 import MediaQuery from 'react-responsive'
 import { ReactClientFieldSettingProps } from '../fieldcontrols/ClientFieldSetting'
 import { reactControlMapper } from '../ReactControlMapper'
-import { RootStateType } from './index'
```
B3zo0: The `RootStateType` is not needed in the index.ts file anymore. Also don't forget to remove the import from sensenet-redux there as well.

PR #195, opened 2019-05-14
WOPI Host page for documents in the browser

herflis:


examples/sn-dms-demo/src/Sensenet.tsx
```diff
@@ -60,6 +60,18 @@ class Sensenet extends React.Component<SensenetProps & ReturnType<typeof mapStat
           }}>
           <HashRouter>
             <Switch>
+              <AuthorizedRoute
+                path="/wopi"
+                authorize={() => this.props.loginState !== LoginState.Unauthenticated}
+                redirectOnUnauthorized="/"
+                render={routerProps => {
+                  const LoadableEditor = Loadable({
+                    loader: () => import(/* webpackChunkName: "dashboard" */ './pages/Editor'),
```
gallayl:
```suggestion
                    loader: () => import(/* webpackChunkName: "editor" */ './pages/Editor'),
```

examples/sn-dms-demo/src/store/wopi/reducers.ts
```diff
@@ -0,0 +1,22 @@
...
+export const wopiData: Reducer<ODataWopiResponse> = (state = wopiDataInitialState, action: AnyAction) => {
+  switch (action.type) {
+    case 'GET_WOPIDATA_SUCCESS':
+      return action.result
+    default:
+      return state
+  }
+}
+
+export const wopi = combineReducers({
```
B3zo0: Is it necessary to call combineReducers with one reducer?

examples/sn-dms-demo/src/store/wopi/reducers.ts
```diff
@@ -0,0 +1,22 @@
...
+const wopiDataInitialState = {
+  accesstoken: '',
+  actionUrl: '',
+  expiration: 0,
+  faviconUrl: '',
+}
+
+export const wopiData: Reducer<ODataWopiResponse> = (state = wopiDataInitialState, action: AnyAction) => {
+  switch (action.type) {
+    case 'GET_WOPIDATA_SUCCESS':
```
B3zo0: I don't see any action with this name. Where does this action come from?

herflis: All the redux actions with ajax request are working this, the ones with the _SUCCES postfix are created in the middleware. :) If you want to catch when the response data is available, you have to use _SUCCESS actions in the reducer. :)

B3zo0: got it! never realized this package `@sensenet/redux-promise-middleware` 😅

examples/sn-dms-demo/src/store/wopi/reducers.ts
herflis: 👍

examples/sn-dms-demo/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -311,7 +319,9 @@ class ActionMenu extends React.Component<
           this.props.history.replace(`/documents?query=${query.Query}&queryName=${query.DisplayName || query.Name}`)
           this.props.closeActionMenu()
           break
-
+        case 'OpenInEditor':
+          return null
+          break
```
B3zo0:
```suggestion
```
break is not needed if there is a return.

B3zo0: Also IMO it would be better to return early in line 173 like this:
```js
      if (!content || action.Name === 'OpenInEditor') {
        return
      }
```

Issue #206, opened 2019-05-21
[FEATURE] Switching between repos should update login window contents


kultsar:
# 🙋 Feature Request

Switching between repos should update login window contents
## 📦 Package / Application

sn-app
## 📃 Version
0.2.0

## 🤔 Expected Behavior
Switching between repos should update login window contents

## 😯 Current Behavior
It just shows last used repo data. Which is fine if I just open the app, but when I switch repo, I want to use that repo, and I want to log in to that repo if I'm logged out of it.

https://www.screencast.com/t/EyGBwAoufnj

## 💁 Possible Solution
Switching between repos should update login window contents

gallayl:
fixed in the referenced PR

kultsar:
Sorry, I was not clear enough. Can we update the user as well?

Issue #205, opened 2019-05-21
[BUG] Switching to repo with expired session should only show login window


kultsar:
# 🐜 Bug Report

Switching to repo with expired session should only show login window and nothing from previous UI. We should not let unauthorized users read data from the screen.

## 📦 Package / Application

sn-app
## 📃 Version
0.2.0

## 🐜 Describe the bug
Switching to repo with expired session should only show login window

## 👉 Steps To Reproduce

1. Log in to 2 different repos, eg. devservice and dmsservice
1. Switch to dmsservice
1. Expire devservice
1. Switch to devservice
1. User should only see login window, and nothing else from previous session (no dashboard, nothing).

## 😯 Current behavior

unauthorized user sees previous ui

## 🤔 Expected behavior

should only see login window

## 🖼 Screenshots

https://www.screencast.com/t/zBYcPGMiSHDD
## 🌍 Your Environment (for developers)

| Software         | Version(s) |
| ---------------- | ---------- |
| Language (js/ts) |
| Operating System |
| NPM/Node/Yarn    |
| Bundler          |

## 🌍 Your environment (for apps / examples)

- Device: laptop
  
- OS: Windows 10
  
- Browser: Chrome
  
- Version: 74.0.3729.157 (Official Build) (64-bit)

gallayl:
The browser won't fire any event if you delete a cookie so the app cannot handle this scenario. The next request will be unauthorized so I think this is not a security issue.

kultsar:
I can accept, but first lets talk, I don't fully understand the situation.

Would it be possible that after user initiates action to switch to a target repository we check if that session expired or not, and if the result is that it expired, we do not show the UI for that repo, just a login screen? I would not like to show anything from the UI (dahsboard, etc) to unauthorized users, if possible. If not, then ok.

gallayl:
1. the next request will kick out the user and she will be forced to refresh the page and log in again
1. if we will have an ordinary dashboard component that fetches info, it will fire that request that will do the trick
1. the placeholder dashboard with Master Roshi doesn't contains any sensitive info

So I don't agree that we should hard code checks on repository switching at the moment.

Issue #204, opened 2019-05-21
[BUG] Typo in failed to login notification message


kultsar:
# 🐜 Bug Report

Typo in failed to login notification message.

## 📦 Package / Application

sn-app

## 📃 Version

0.2.0
## 🐜 Describe the bug

An apostrophe is missing after the username in the notification message that is displayed after failed login attempts:

```Failed to log in with user 'businesscat@sensenet.com to repository 'https://devservice.demo.sensenet.com```

## 👉 Steps To Reproduce

1. Reproduce a failed login (eg. wrong password) to a repo.
1. Check notification error message

## 😯 Current behavior

Apostrophe missing. Immediate disappointment ensues, followed by permanent despair.

## 🤔 Expected behavior

The apostrophe should not be missing. This world needs more apostrophes!!

## 🖼 Screenshots

![image](https://user-images.githubusercontent.com/39081660/58077222-56a8b880-7bac-11e9-8ad4-e605007e2c8b.png)

## 🌍 Your Environment (for developers)

| Software         | Version(s) |
| ---------------- | ---------- |
| Language (js/ts) |
| Operating System |
| NPM/Node/Yarn    |
| Bundler          |

## 🌍 Your environment (for apps / examples)

- Device: laptop
  
- OS: Windows 10
  
- Browser: Chrome
  
- Version: 74.0.3729.157 (Official Build) (64-bit)

gallayl:
fixed in the referenced PR

kultsar:
Apostrophe still missing after username in login notification message for me when logging in at https://5cebd0a75ff1ede070194da2--sn-app-dev.netlify.com

gallayl:
![image](https://user-images.githubusercontent.com/16716099/58476879-60a15d00-8152-11e9-91ce-8744ebea81fc.png)

kultsar:
My bad, mixed up with successful login message.

Issue #203, opened 2019-05-20
[BUG] Failed login to invalid url repo prevents login to valid url repo


kultsar:
# 🐜 Bug Report

Failed login to bad url repo prevents login to good url repo
## 🐜 Describe the bug

A previous failed login to bad url repo prevents login to good url repo.

## 👉 Steps To Reproduce

1. Log in with valid url repo
1. Log out
1. Try to log in with invalid url repo
1. Try to log in with valid url repo

## 😯 Current behavior

Cannot login. Error message refers to bad url repo. Until user reloads the page, cannot log in to any valid url repo.

## 🤔 Expected behavior

Should be able to log in.

## 🖼 Screenshots

https://www.screencast.com/t/seUPJC02js

## 🌍 Your Environment (for developers)

| Software         | Version(s) |
| ---------------- | ---------- |
| Language (js/ts) |
| Operating System |
| NPM/Node/Yarn    |
| Bundler          |

## 🌍 Your environment (for apps / examples)

- Device: laptop
  
- OS: Windows 10
  
- Browser: Chrome
  
- Version: 74.0.3729.157 (Official Build) (64-bit)

gallayl:
fixed in the referenced PR

PR #245, opened 2019-05-27
[admin-ui]: fixes in login and repo switching

gallayl:
### Changes
 - Each and every component uses the *real* session state from now. Removed the debounced session state
 - Welcome message and animation will be displayed **only** for the dedicated login routes (e.g.: **/#/login** or **/#/:repo/login**
 - LoginComponent will update the repository URL and username when you switch between repositories. Removed the initialization with defaultValue.
 - Repository URL will be a required field

Related issues: 
 #199 #200 #203 #204 #205 #206


kultsar:
All other issues are ok and tested, but we need to update user as well in the login window (https://github.com/SenseNet/sn-client/issues/206#issuecomment-496511194)

PR #271, opened 2019-06-03
Feature/materialv4

B3zo0:
Tasks moved from issue:

## Common tasks:
 - [x] Investigate and fix all breaking changes (needs more info). `# fixed all type errors and changes`
 - [x] We should check / investigate how we can use the native tree shaking with **destructured imports** (see the linked article). We should merge / simplify our imports. - `# I couldn't make this work. I tried with dms demo. didn't see any change in bundle size. Tried with search-react as well. When not used default export then it pulled dependencies that we didn't used there like Tabs. `

## Packages
 - [x] controls-react
   - [x] Also check the compatibility with 3rd party pickers (e.g. date, time) - `# upgraded the picker to 3.0.0 as well.`
 - [x] icons
 - [x] pickers-react
 - [x] list-controls-react
 - [x] search-react
 - [x] document-viewer-react

## Apps

 - [x] DMS
   - [x] Keep an eye on the bundle size with bundle-size-analyzer 👀
 - [x] Admin UI: 
   - [x] Check react-autosuggest compatibility (command palette)
   - [x] We use an own ThemeContext implementation. Check for built-in alternative.
   - [x] Keep an eye on the bundle size with bundle-size-analyzer 👀
 - [x] Storybook


herflis:
- [ ] padding in actionmenus (dropdowns) between icon and text
image
- [ ] missing paddings (top and bottom) in dialogs
image
- [ ] (I'm not sure that it is because of the upgrade) but the overlay icons are mispositioned
image

gallayl:
The table (maybe the ContentList(?)) has a huge additional padding
image

kultsar:
- [ ] Approve/reject fell apart
image

- [ ] Versions look strange for me
image

kultsar:
- [ ] Field controls/Color picker does not work. I click in the field, but nothing happens.

gallayl:
The Color Picker doesn't seem to work :( 
https://5cf654df8db256d3f9cd1142--sn-react-component-docs-dev.netlify.com/?path=/story/fieldcontrols-colorpicker--new-mode

Issue #290, opened 2019-06-11
[controls-react] Tests and accesibility fixes


herflis:
- [ ] Check the 'data-x' params and rename them if it is possible (remove the 'data-' prefix)
- [ ] Write tests for testing the controls (it is not a must to reach 100% coverage ofc)
- [ ] Storybook mentions some accessibility issues with the controls, if it is not a big effort solve them now

PR #297, opened 2019-06-17
Feature/data props rename

B3zo0:
## What I did in this pr:
_This pr is part of https://github.com/sensenet/sn-client/issues/290_
- Added new eslint rule [sort-imports](https://eslint.org/docs/rules/sort-imports)
  - Fixes included
- Turned off jsdocs comments in react-components-docs
- Added fileName prop to UploadOptions, this way you can overwrite a content's binary if you pass the name of the content
- ControlMapper now returns the component that needs to be rendered
- ReactControlMapper is only mapping the field settings with the components now
  - Added currency field setting and it maps to number control
  - Binary field setting to map to fileUpload control
- Controls
  - Flatened structure. Moved the controls that only have one component to its parent folder
  - Changed ReactFieldSettings to use the fieldsetting related to the control
  - Removed all the dot notations
  - Unified the new/edit, browse/default views in almost every component
  - Removed Radium from components
  - Unified DatePicker and DateTimePicker it is one component now and renders based on DateTimeMode setting
  - Removed DisplayName control
- Default content types
  - Changed the TextType from enum to string literal
  - Added ColorFieldSetting, CurrencyFieldSetting
- Component docs
  - Moved the custom schema to a new file from ViewControl stories
  - Field control stories are generated dynamically now

Remaining tasks:

- [x] Fix Todos
- [x] Add missing props to fieldcontrols
- [x] Fix control-mapper tests
- [x] File upload control as attachment
- [x] Remove radium from package.json
- [x] FileName control extension prop
- [x] ShortText MinLength, MaxLength
- [x] EditView, NewView remove comments


packages/sn-controls-react/src/fieldcontrols/ClientFieldSetting.ts
```diff
@@ -1,88 +1,27 @@
...
-  'data-actionName'?: 'new' | 'edit' | 'browse'
-  /**
-   * Text of the hint that could be displayed after the field control
-   */
-  'data-hintText'?: string
-  /**
-   * Text of the placeholder
-   */
-  'data-placeHolderText'?: string
+  actionName?: 'new' | 'edit' | 'browse'
```
B3zo0: This type should come from default-content-types.

packages/sn-controls-react/src/fieldcontrols/ClientFieldSetting.ts
```diff
@@ -1,88 +1,27 @@
...
-  /**
-   * Name of the fieldcontrol type
-   */
-  'data-typeName'?: string
-  /**
-   * Url of the repository
+   * Function for rendering an icon by an iconname given as an input prop
    */
-  'data-repositoryUrl'?: string
+  renderIcon?: (name: string) => JSX.Element
```
B3zo0:
```suggestion
  renderIcon?: (name: string) => React.ReactElement
```

packages/sn-controls-react/src/fieldcontrols/Avatar/Avatar.tsx
```diff
@@ -219,49 +167,42 @@ export class Avatar<T extends GenericContent, K extends keyof T> extends Compone
...
                 </Typography>
                 <AvatarPicker
-                  path={
-                    this.props['data-selectionRoot']
-                      ? this.props['data-selectionRoot'][0]
-                      : this.props['data-uploadFolderPath']
-                  }
-                  allowedTypes={this.props['data-allowedTypes']}
-                  repository={this.props['data-repository']}
+                  path={this.props.settings.SelectionRoots ? this.props.settings.SelectionRoots[0] : ''}
```
herflis: There should be a default url instead of an empty string. In this case the uploadFolderPath gave the possibility to add a default path where the avatar image should be uploaded if there's no dedicated place for the avatars and the selection roots is empty. For example if I want to force users to upload their avatar into their own profile doclib or something like that which cannot be defined in the selectionroot

packages/sn-controls-react/src/fieldcontrols/CheckboxGroup.tsx
```diff
@@ -0,0 +1,161 @@
...
+import FormControl from '@material-ui/core/FormControl'
+import FormControlLabel from '@material-ui/core/FormControlLabel'
+import FormGroup from '@material-ui/core/FormGroup'
+import FormHelperText from '@material-ui/core/FormHelperText'
+import FormLabel from '@material-ui/core/FormLabel'
+import TextField from '@material-ui/core/TextField'
+import { ChoiceFieldSetting } from '@sensenet/default-content-types'
+import { ReactClientFieldSetting } from './ClientFieldSetting'
+
+// TODO(Zoli): Review this. It's value should be determind by the allowmultiple setting
```
herflis: @B3zo0 I hope you've checked this 😃

packages/sn-controls-react/src/fieldcontrols/Number.tsx
```diff
@@ -0,0 +1,109 @@
...
+      this.props.settings.Type === 'IntegerFieldSetting'
+      ? 1
+      : 0.1
+  }
+
+  /**
+   * Returns inputadornment by currency and percentage settings
+   */
+  public defineCurrency = () => {
+    // Todo: Currency
```
herflis: ☝️

B3zo0: I've tried to do that but couldn't figure out the proper way. Sensenet only provides a Format property which is a culture code. Not really helpful. Not sure how this should be implemented. Any thoughts?

packages/sn-controls-react/src/fieldcontrols/CheckboxGroup.tsx
B3zo0: Good catch. I've reviewed and reworked this component.

packages/sn-controls-react/src/fieldcontrols/Avatar/Avatar.tsx
B3zo0: I've put back uploadFolderPath to this control.

packages/sn-controls-react/src/fieldcontrols/type-guards.ts
```diff
@@ -0,0 +1,10 @@
+import { GenericContent, User } from '@sensenet/default-content-types'
+
+/**
+ * Typeguard for user. let typescript know that the content is a User.
```
herflis:
```suggestion
 * Typeguard for user. Lets typescript know that the content is a User.
```

PR #322, opened 2019-07-09
[admin-ui]: Content Type editing fixes

gallayl:
The CurrentAncestors provider has broke the app if it could not load the ancestors list (e.g. editing a content type)
This PR changes this behavior, if the custom action fails:
 - [x] We won't throwi up the error
 - [x] Add a warning entry to the log with the related content and the error details
 - [x] The Ancestors list will contain only the Current Content

### DIFF
--- a/apps/sensenet/src/context/CurrentAncestors.tsx
+++ b/apps/sensenet/src/context/CurrentAncestors.tsx
@@ -3,7 +3,7 @@ import { debounce } from '@sensenet/client-utils'
 import { GenericContent } from '@sensenet/default-content-types'
 import React, { useContext, useEffect, useState } from 'react'
 import Semaphore from 'semaphore-async-await'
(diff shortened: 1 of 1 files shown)


apps/sensenet/src/context/CurrentAncestors.tsx
```diff
@@ -78,7 +80,10 @@ export const CurrentAncestorsProvider: React.FunctionComponent = props => {
   }, [currentContent.Id, loadLock, reloadToken, repo])
 
   if (error) {
-    throw error
+    logger.warning({
+      message: `Error loading ancestors. ${error.toString()}`,
```
B3zo0: Imo it would be better to show only the error message here. all other things are passed in the details.
```suggestion
      message: `Error loading ancestors. ${error.message}`,
```

PR #340, opened 2019-08-02
[admin-ui] Drawer customization and refactor

gallayl:
### Features
- [x] Content views (simple, explorer, commander) and Breadcrumbs now have an optional Root property

### Breaking changes
 - [x] The Browse and Search url schema has been changed from ``/:repo/browse/:contentId?`` to ``/:repo/browse/:browseData?``. Browse data contains a json-encoded structure of the current browse type, root, fields to display, load settings, etc.
 - [x] The Drawer settings's **items** type has been changed from a enum array (``["search", "content"]``) to an object array (``[ { "itemType": "Content", "settings": { "root": "/Root/Sites", "title": "MySetting", "icon": "Resource"} }]``). **We should discuss how can we migrate previous settings**
 - [x] - Drawer settings will be reseted if there are any string values in there (old schema)
----
### Misc, refactor & optimizations
 - [x] Moved ContentList to a separate directory and extracted custom field components
 - [x] Actions field in the Content List now uses the unified context menu instead of separated instance per row (was a performance issue)
 - [x] Icons now contains a "*wellKnownIconNames*" tuple (used by intellisense for drawer icons)
 - [x] Unified content navigation logic from components (explorer, commander, simple)
 - [x] Icons - Unified schema resolver


herflis:
@gallayl I've checked the admin-ui visually and I've found the following issues:
- [ ] sometimes the query results are messed up, for example I choose users and groups, it loads the users and groups, then I choose content types and the result list is not cleared, both the users, groups and the contentypes are listed
image
- [ ] we should hide the breadcrumb somehow on places where the user cannot navigate up normally, e.g. from the ctd or localization sources
image
- [ ] date values in the content list should be formatted (e.g. '2 minutes ago' instead of 2019-08-06T09:30:56.077Z)
- [ ] **the last one is for me** because I do not want to forget it: we should do something with the local groups in the users and groups list

PR #360, opened 2019-08-21
ci: github actions start

B3zo0:


B3zo0:
This looks promising but without cache it is still seems slow.

PR #383, opened 2019-09-04
Viewer Hooks ⚓

gallayl:
- [x] Refactor the state logic from Redux to Hooks
   - [x] Refactor all components from Class Components to FC-s
   - [x] Class state to useState() hook
   - [x] Redux states to Context and ContextProvider (use the Hooks where possible)
   - [x] Clean up ``mapStateToProps``, ``mapDispatchToProps``, fix prop types
   - [x] Remove react-redux ``connect`` HOC, adjust exports and export names
   - [x] remove redux store, action creators, dispatchers, etc...
 - [x] Fix in Storybook
 - [x] Fix in the Admin-UI
 - [x] Fix DMS
 - [x] Fix tests
 - [x] Clean up dependencies
 - [x] Document changes and breaks, add missing code docs on public API

# Breaking change 💥
- All component in the viewer are now function components and using hooks and context api
- Redux is removed from the library. It is using context api now for state management.
- Cleaner and easier to use api. No need to define your own implementation of the endpoints if not needed.
- Built in defaults for api endpoints/localization/theme
- See usage example in components-docs


packages/sn-document-viewer-react/src/context/document-data.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }, [api, doc.documentIdOrPath, repo.configuration.repositoryUrl, doc.version])
+
+  const updateDocumentData = useCallback(
+    async (newDocData: Partial<DocumentData>) => {
+      const merged = deepMerge(documentData, newDocData)
+      if (JSON.stringify(documentData) !== JSON.stringify(merged)) {
+        try {
+          await loadLock.acquire()
+          setDocumentData(merged)
+          console.log(documentData, merged)
```
herflis: is this console.log necessary here?

packages/sn-document-viewer-react/src/hooks/usePreviewImage.ts
```diff
@@ -0,0 +1,18 @@
+import { useCallback } from 'react'
+import { usePreviewImages } from './usePreviewImages'
+
+export const usePreviewImage = (pageNo: number) => {
+  const images = usePreviewImages()
+  const image = images.imageData.find(i => i.Index === pageNo)
+  const { imageData, ...context } = { ...images }
+
+  // ToDo: Polling?
```
herflis: do we need to talk about this?

packages/sn-document-viewer-react/test/DocumentViewerLayout.test.tsx
```diff
@@ -17,153 +22,127 @@ declare global {
...
   })
 
-  it('should scroll to page when active pages changed', () => {
-    const setActivePages = jest.fn()
-    const wrapper = shallow(
-      <DocumentViewerLayoutComponent {...defaultProps} setActivePages={setActivePages} showThumbnails={false}>
-        {'some children'}
-      </DocumentViewerLayoutComponent>,
+  /**
+   * This test doesn't make any sense to me (Zoli)
```
herflis: so why is it here? 😃

packages/sn-document-viewer-react/src/hooks/usePreviewImage.ts
herflis: @gallayl can you please help us solve this riddle? 😄

gallayl: 🃏 I think the issue was that if you have only a few preview images generated (e.g. the first 10 only) but you want to see the 11th, you have to start a similar polling as for document initialization but with the exact page Id. I'm not sure that this logic was moved from Redux to another hook but I think it should be here :)

B3zo0: Thanks, I've fixed that.

PR #393, opened 2019-09-19
Dependency update - 2019 September 🍁

gallayl:
- [x] Updated dependencies
 - [x] Fixed breaks
 - [x] Additional type castings
----
As react-async has a very little benefit (~5 lines of code vs an additional dependency), we've agreed to replace with useEffect() during the dependency update to fix a typing issue.
 - [x] Refactor the one occurence in Pickers
 - [x] Remove from dependencies
 - [x] Fix ``useAsync()`` mocks in tests
   - [x] Pickers
   - [x] Avatar
   - [x] ReferenceList


packages/sn-client-utils/src/index.ts
```diff
@@ -10,3 +10,5 @@ export { sleepAsync } from './SleepAsync'
 export { Trace } from './Trace'
 export { toNumber } from './ToNumber'
 export { tuple } from './Tuple'
+export * from './AsyncReturnValue'
+export * from './PromiseReturnValue'
```
B3zo0: I'm not sure if these types are needed.

gallayl: There are only one usage here: https://github.com/SenseNet/sn-client/pull/393/files#diff-240e5c83bc40b7b5b3fe234adbd6d05eR59
We can remove them, but then we have to use an inline type / explicit type here.

PR #412, opened 2019-10-08
Bug fix #401

### LINKED ISSUE #401: [BUG] Cannot save a query
# 🐜 Bug Report

I tried to save a query (search result) but it does not do anything when I hit the save button.

## 📦 Package / Application

@sensenet/sn-app

## 📃 Version

latest

## 🐜 Describe the bug

If i hit the save button on the modal window it does not do anything. On the console I can see some errors related to the missing user profiles, since the action is try to save the non-public queries into the current users user profile.

Part of the issue cannot be solved since the user profile feature is not the part of the latest sensenet service. 

## 👉 Steps To Reproduce

1. Go to [admin.sensenet.com](https://admin.sensenet.com)
2. Login with Builtin\admin password: admin, repositoryurl: https://dev.demo.sensenet.com
3. Write the query into the command palette ```+TypeIs: Article``` and choose 'Open in search'
4. Click on save query
5. Push the save button

## 😯 Current behavior

Nothing happens, no save, no modal close...nothing

## 🤔 Expected behavior

I think since the user profile feature is not in the sensenet service package and the demo install, 'public' checkbox is unnecessery on the modal. The queries could be saved only as public queries so it should by set by default to true. As I see now, even if I choose to save the query as public it does nothing so, this part of the issue should be also investigated and fixed.

[here](https://github.com/SenseNet/sn-client/blob/master/apps/sensenet/src/components/search/index.tsx#L197-L200) the FormLabel could be removed completely and [here](https://github.com/SenseNet/sn-client/blob/master/apps/sensenet/src/components/search/index.tsx#L207) the action should be called on the current content /context not on the current users profile path (the ```SaveQuery``` action choose where to save it by the ```queryType``` param).

atomanonymous:
made the necessary changes


B3zo0:
Hi @atomanonymous, thanks for the fix. Can you please remove the checkbox mentioned in the issue as well? Also please fix your lint issues.

Issue #403, opened 2019-10-03
Edit sn-react-typescript-boilerplate to use sn-hooks-react


B3zo0:
Once #402 is done, we should rewrite it to use sn-hooks-react.

- [x] Replace useRepository with useRepository from `@sensenet/hooks-react`
- [ ] Add `@sensenet/hooks-react` to package.json and to the tsconfig.json

Ishaan28malik:
Hey I would like to work on this @B3zo0 
But not sure where to implement this change?

B3zo0:
Fixed in #415

PR #415, opened 2019-10-08
Fix issue #403

dlemstra:
This PR fixes the issue reported in #403. One of the requested changes was an update to `tsconfig.json` but I don't understand why that file should be changed.


B3zo0:
Hi @dlemstra thank you for your contribution. The idea behind the tsconfig.json update was to add the `@sensenet/hooks-react` package as a project reference. We can skipp that now because on order to make it work we should change the webpack config as well. Don't forget to fill out our [form](https://www.sensenet.com/Hacktoberfest2019) to get a 👕

PR #432, opened 2019-10-17
moved logging and inject to client-utils

SachinKumar105:


B3zo0:
While I appreciate that you are trying to help this pr is wrong in so many levels. Also there is another guy that is currently working on this. Thank you for your time.

PR #433, opened 2019-10-17
added "experimentalDecorators": true to tsconfig.json file

SachinKumar105:


B3zo0:
Hi @SachinKumar105, I'm sorry but that is not something that we are looking for right now.

PR #454, opened 2019-11-06
Feature/context menu

B3zo0:
- [x] Add hard coded strings to localization
- [x] Wopi action should only show one (edit/see)
- [x] Review and open tasks (issues) for todo comments
- [x] Use dialog provider for all dialogs

There was a Todo about checking mobile view. This is messed up now. It does not work as expected. We should review mobile view once we have proper design for admin-ui.

## Changes
- eslint rule require-jsdoc is disabled for apps and examples. (I found it not helpful during development. Also IMO examples and apps do not need code comments strictly for every function.)
- 🔥 DialogProvider - there is a new hook called useDialog that helps managing them. Every dialog is going through that for now and every dialog is loaded lazily.
  - every dialog that was in the app is now refactored out to dialogs folder and uses dialog provider
- Upload dialog is removed from it's own route. It felt weird in the first place. The functionality is not changed. Only thing is that it wont close the dialog window when we push browser back.
- Error boundary refactored to be a generic component so it can be used more than one time. It accepts a FallbackComponent to show when error is caught.
- Content list component has a Locked field now that shows if the content is checked out to someone or not
- Because set permission action is not yet implemented it is set to disabled in the context menu
- 🌟 New dialogs:
  - Approve, reject
  - Versions
  - Check in
  - Are you sure (generic component)
- ODataUrlBuilder fix -> it is not mutating any params now.
- 💥 useWopi hook is refactored to return two functions instead of values
  - example usage: 
```ts
const { isWriteAvailable } = useWopi()

isWriteAvailable(content) && someFunction()
```


apps/sensenet/src/components/content-list/locked-field.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return checkedOutTo.Name === session.currentUser.Name ? 'Me' : checkedOutTo.FullName ?? 'Someone'
+  }
+
+  if (!content.Locked && !content.Approvable) {
+    // We need to return an empty TableCell so the Table remains aligned.
+    return <TableCell />
+  }
+
+  return (
+    <TableCell>
```
herflis:
```suggestion
    <TableCell align="center">
```

apps/sensenet/src/components/content-list/locked-field.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <Tooltip title={content.Approvable ? localization.actionNeeded : localization.checkedOutTo(lockedByName())}>
+        {content.Approvable ? <AssignmentIcon /> : <Lock />}
```
herflis: I would change the assignment to assignment_late

apps/sensenet/src/localization/default.ts
```diff
@@ -51,6 +45,29 @@ const values = {
     created: 'Created',
     unknownOwner: 'Unknown',
   },
+  checkInDialog: {
+    inputLabel: 'Check in comment (optional)',
+    successMessage: 'Check in succeded',
+    errorMessage: 'Check in failed',
```
herflis:
```suggestion
    errorMessage: 'Check in failed',
    send: 'Send'
```

apps/sensenet/src/components/dialogs/check-in.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <TextField
+          label={localization.inputLabel}
+          multiline
+          onChange={event => setComment(event.target.value)}
+          value={comment}
+          fullWidth
+        />
+      </DialogContent>
+      <DialogActions>
+        <Button onClick={onSubmit}>Send</Button>
```
herflis:
```suggestion
        <Button onClick={onSubmit}>{localization.send}</Button>
```

apps/sensenet/src/components/dialogs/check-in.tsx
```diff
@@ -0,0 +1,53 @@
...
+    } catch (error) {
+      logger.warning({ message: localization.errorMessage, data: error })
+    } finally {
+      closeLastDialog()
+    }
+  }
+
+  return (
+    <>
+      <DialogTitle>Add a check in comment</DialogTitle>
```
herflis:
```suggestion
      <DialogTitle>{localization.checkinComent}</DialogTitle>
```

apps/sensenet/src/localization/default.ts
```diff
@@ -380,6 +399,21 @@ const values = {
...
+    commentTableHead: 'Comment',
+    rejectReasonTableHead: 'Reject reason',
+    restoreTableHead: 'Restore to version',
+    restoreBodyText: (name: string, version?: string) =>
+      `Are you sure you want to restore version <strong>${version}</strong> of <strong>${name}</strong>`,
+    restoreSubmitText: 'Restore',
+    restoreVersionSuccess: (name: string, version?: string) => `${name} restored to version ${version}`,
+    restoreVersionError: (name: string, version?: string) =>
+      `Couldn't restore version to  ${version} for content: ${name}`,
+  },
```
herflis:
```suggestion
restoreButtonTitle: 'Restore version'
  },
```

apps/sensenet/src/components/dialogs/versions.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  </Tooltip>
+                </TableCell>
+                <TableCell>
+                  <Tooltip disableFocusListener={true} title={version.RejectReason ?? ''}>
+                    <span>{version.RejectReason ?? ''}</span>
+                  </Tooltip>
+                </TableCell>
+                <TableCell padding="none" style={{ width: '5%' }}>
+                  {index !== versions.length - 1 ? (
+                    <IconButton onClick={() => restoreVersion(version)} title="Restore version">
```
herflis:
```suggestion
                    <IconButton onClick={() => restoreVersion(version)} title="{localization.restoreButtonTitle}">
```

apps/sensenet/src/localization/hungarian.ts
```diff
@@ -82,8 +75,9 @@ const values: DeepPartial<typeof import('./default').default> = {
     logoutCancel: 'Mégsem',
     logoutButtonTitle: 'Kijelentkezés',
     logoutDialogTitle: 'Biztosan kijelentkezel?',
-    loggingOutFrom: "Kijelentkezés a '{0}' repository-ból... ",
-    logoutConfirmText: 'Jelenleg a {0} repository-t használod {1} felhasználóval. Biztosan kijelentkezel?',
+    loggingOutFrom: (repoUrl: string) => `Kijelentkezés a ${repoUrl} repository-ból...`,
```
herflis: Some of the string resources added to the default.ts are completely missing from the hungarian.ts

apps/sensenet/src/components/content-list/locked-field.tsx
B3zo0: Ok, I've changed it to assignment_late.

PR #460, opened 2019-11-15
Fix 'undefined' DisplayName in notification after editing a content

herflis:


packages/sn-controls-react/src/viewcontrols/EditView.tsx
```diff
@@ -41,89 +30,66 @@ export interface EditViewState {
...
-    }
+export const EditView: React.FC<EditViewProps> = props => {
+  const controlMapper = reactControlMapper(props.repository)
+  const schema = controlMapper.getFullSchemaForContentType(props.content.Type, 'edit')
+  const [saveableFields, setSaveableFields] = useState({})
 
-    this.handleInputChange = this.handleInputChange.bind(this)
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
+    props.onSubmit && props.onSubmit(props.content, saveableFields as any)
```
B3zo0:
```suggestion
    props.onSubmit?.(props.content, saveableFields as any)
```
With the new TypeScript 3.7 we no longer need to write code like this: `object.function && object.function()`
we can simplify this with this syntax `object.function?.()` . This will not execute the function if it is null or undefined.

packages/sn-controls-react/src/viewcontrols/EditView.tsx
```diff
@@ -41,89 +30,66 @@ export interface EditViewState {
...
+    props.submitCallback && props.submitCallback()
```
B3zo0:
```suggestion
    props.submitCallback?.()
```

packages/sn-controls-react/src/viewcontrols/EditView.tsx
```diff
@@ -41,89 +30,66 @@ export interface EditViewState {
...
   }
 
-  /**
-   * handle change event on an input
-   */
-  public handleInputChange(field: string, value: any) {
-    this.setState({
-      content: { ...this.state.content, [field]: value },
```
B3zo0: I wonder why did we update the content here? Thoughts?

packages/sn-controls-react/src/viewcontrols/EditView.tsx
```diff
@@ -4,34 +4,23 @@
...
 import MediaQuery from 'react-responsive'
 import { reactControlMapper } from '../ReactControlMapper'
-import { ReactClientFieldSetting } from '../fieldcontrols/ClientFieldSetting'
 
 /**
  * Interface for EditView properties
  */
 export interface EditViewProps {
   content: ContentType
-  onSubmit?: (id: number, content: GenericContent) => void
```
B3zo0: This was kind of a lie in the first place. We return fields that changed not a GenericContent type. People might think that there is a Name, Id, Path field in this object but Id will never be in the changed fields. We should think about this in the future.

herflis: saveableFields' type is now changed to `Partial<GenericContent>`

PR #465, opened 2019-11-20
Fix link to the getting started docs

herflis:


apps/sensenet/src/services/PersonalSettings.ts
```diff
@@ -255,7 +255,7 @@ export const defaultSettings: PersonalSettingsType = {
         widgetType: 'markdown',
         settings: {
           content:
-            '<div>To get started with sensenet</div><br /><a style="color: #26a69a; line-height: 2rem" href="https://community.sensenet.com/docs/concepts/overview">Overview</a><br /><a style="color: #26a69a; line-height: 2rem" href="https://community.sensenet.com/docs/tutorials/getting-started">Getting started</a><br /><a href="https://community.sensenet.com/tutorials" style="color: #26a69a; line-height: 2rem">Tutorials</a><br /><a style="color: #26a69a; line-height: 2rem" href="https://community.sensenet.com/docs/examples">Example apps</a>',
+            '<div>To get started with sensenet</div><br /><a style="color: #26a69a; line-height: 2rem" href="https://community.sensenet.com/docs/concepts/overview">Overview</a><br /><a style="color: #26a69a; line-height: 2rem" href="https://community.sensenet.com/docs/getting-started">Getting started</a><br /><a href="https://community.sensenet.com/tutorials" style="color: #26a69a; line-height: 2rem">Tutorials</a><br /><a style="color: #26a69a; line-height: 2rem" href="https://community.sensenet.com/docs/examples">Example apps</a>',
```
B3zo0: I think it would make sense to open these links in new tab.

herflis: 👍

B3zo0:
I see 404 for [Overview](https://community.sensenet.com/docs/concepts/overview) this could be https://community.sensenet.com/docs/concepts maybe? also 404 for example apps.

herflis:
@B3zo0 I've changed the link you've mentioned first to the overview on  sensenet.com and till the example apps are not the part of the monorepo, the related article is not published so I've changed the link to a '\'.

PR #470, opened 2019-11-25
Personal settings model update

B3zo0:
Updated the model for drawer item permission from role to action. We don't use role property.

Added some more descriptions and fixed some mistakes.

### DIFF
--- a/apps/sensenet/src/services/MonacoModels/PersonalSettingsModel.ts
+++ b/apps/sensenet/src/services/MonacoModels/PersonalSettingsModel.ts
@@ -273,13 +273,19 @@ export const setupModel = (language = defaultLanguage, repo: Repository) => {
                             settings: {
                               type: 'object',
                               properties: {
-                                title: { type: 'string' },
-                                description: { type: 'string' },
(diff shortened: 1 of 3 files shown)


apps/sensenet/src/services/MonacoModels/PersonalSettingsModel.ts
```diff
@@ -247,8 +247,8 @@ export const setupModel = (language = defaultLanguage, repo: Repository) => {
                             settings: {
                               type: 'object',
                               properties: {
-                                term: { type: 'string', description: language.drawer.contentRootDescription },
-                                columns: { $ref: '#/definitions/columns' },
+                                term: { type: 'string', description: language.drawer.queryTerm },
+                                columns: { $ref: '#/definitions/columns', description: language.drawer.columns },
```
pusztaienike: This line can causes the duplicated description on hover.  ?!

B3zo0: Indeed. Good catch! Thanks 😊

PR #485, opened 2019-12-02
Various fixes

herflis:
- Fix: creating a new content with a ContentTemplate
- Fix: password field's value is now saved when a new user is created
- Fix: tagsinput query with multiple allowed types
- Fix: adding and deleting reference in the tagsinput


packages/sn-controls-react/src/fieldcontrols/Password.tsx
```diff
@@ -9,15 +9,21 @@ import Input from '@material-ui/core/Input'
...
 /**
  * Field control that represents a Password field. Available values will be populated from the FieldSettings.
  */
 export const Password: React.FC<ReactClientFieldSetting> = props => {
+  const [value, setValue] = useState(props.fieldValue || changeJScriptValue(props.settings.DefaultValue) || '')
   const [showPassword, setShowPassword] = useState(false)
 
+  const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement | HTMLSelectElement>) => {
+    setValue(e.target.value)
+    props.fieldOnChange && props.fieldOnChange(props.settings.Name, e.target.value)
```
B3zo0:
```suggestion
    props.fieldOnChange?.(props.settings.Name, e.target.value)
```

packages/sn-controls-react/src/viewcontrols/NewView.tsx
```diff
@@ -14,7 +14,7 @@ import { reactControlMapper } from '../ReactControlMapper'
  * Interface for NewView properties
  */
 export interface NewViewProps {
-  onSubmit?: (path: string, content: ContentType, contentTypeName: string) => void
+  onSubmit?: (path: string, content: ContentType, contentTypeName: string, contentTemplateName?: string) => void
```
B3zo0: I wonder if it is really needed to pass the contentTypeName twice here?

herflis: What if a user want to pass not the contenttype's name but a custom contenttemplate?

packages/sn-controls-react/src/fieldcontrols/Password.tsx
```diff
@@ -9,15 +9,21 @@ import Input from '@material-ui/core/Input'
...
 import Typography from '@material-ui/core/Typography'
+import { changeJScriptValue } from '../helpers'
 import { renderIconDefault } from './icon'
 import { ReactClientFieldSetting } from './ClientFieldSetting'
 
 /**
  * Field control that represents a Password field. Available values will be populated from the FieldSettings.
  */
 export const Password: React.FC<ReactClientFieldSetting> = props => {
+  const [value, setValue] = useState(props.fieldValue || changeJScriptValue(props.settings.DefaultValue) || '')
```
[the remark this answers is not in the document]
herflis: I've changed the mentioned part and also made the password field hidden on the edit and browse views in the default schema

packages/sn-controls-react/src/viewcontrols/NewView.tsx
B3zo0: Because we pass in the `schema.ContentTypeName`, the user should handle itself. I still think that passing this down in this component is not necessary.

Issue #491, opened 2019-12-03
[BUG] Cannot select the same file right after you select the file for upload but cancelled it


pusztaienike:
# 🐜 Bug Report

Cannot select the same file right after you select the file for upload but cancelled it.

## 📦 Package / Application

@sensenet/sn-app

## 📃 Version

latest

## 👉 Steps To Reproduce

1. Go to admin.sensenet.com and login with admin
2. Create a DemoWorkspace into /Content
3. Add a folder to the demoworkspace
4. Select an image to upload
5. Click the close icon ( X ) which located in the row where the selected image is
6. Select the same image again
Result: The selected file is unvisible

## 😯 Current behavior

Result: The selected file is unvisible

## 🤔 Expected behavior

Expected: The selected file is visible, and it can be uploaded

## 🗨 Additional context

I tried to upload an image file (.jpg)

PR #495, opened 2019-12-03
Set ev.target.files empty after every change

pusztaienike:
If you select a file for upload it will be stored in ev.target.files, but if you cancel the upload and select the same file again onChange won't be called [- cause your input will be the same as in ev.target.files-] . 
My solution was to set event.target.value to an empty string which set the event.target.files empty as well.

### DIFF
--- a/apps/sensenet/src/components/dialogs/upload/upload-dialog.tsx
+++ b/apps/sensenet/src/components/dialogs/upload/upload-dialog.tsx
@@ -211,7 +211,10 @@ export function UploadDialog(props: UploadDialogProps) {
         </Grid>
       </DialogContent>
       <input
-        onChange={ev => ev.target.files && addFiles([...ev.target.files])}
+        onChange={ev => {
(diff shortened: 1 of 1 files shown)


apps/sensenet/src/components/dialogs/upload/upload-dialog.tsx
```diff
@@ -211,7 +211,10 @@ export function UploadDialog(props: UploadDialogProps) {
         </Grid>
       </DialogContent>
       <input
-        onChange={ev => ev.target.files && addFiles([...ev.target.files])}
+        onChange={ev => {
+          ev.target.files && addFiles([...ev.target.files])
+          ev.target.value = ''
```
B3zo0: I would rather see this fix in the removeItem method. Set `inputFile` value field and add a comment please that states why was this necessary.

pusztaienike:
This closes #491

Issue #489, opened 2019-12-03
[BUG] Schema is not automatically updated on ctd change


herflis:
# 🐜 Bug Report

If user saves a CTD the underlying schema does not refresh only if user reloads the app.

## 👉 Steps To Reproduce

1. Login to admin.sensenet.com with the builtin admin user
2. Open the ctd list with clicking the related menuitem in the drawer
3. Open type Memo for editing
4.  Add `<VisibleEdit>Hide</VisibleEdit>` to MemoType field's Configurations and save the ctd
5. Open a Memo for edit

![image](https://user-images.githubusercontent.com/6968860/70032289-7bdfa000-15ad-11ea-8fe4-129d381921b3.png)

6. Reload the app with F5 and open the same Memo for edit

![image](https://user-images.githubusercontent.com/6968860/70032396-b2b5b600-15ad-11ea-9122-24e228604d69.png)

PR #501, opened 2019-12-05
Fix/automatically updated schema

pusztaienike:
The mentioned bug has been fixed, and the reload on the binary editor also


pusztaienike:
This closes #489

Issue #425, opened 2019-10-10
[admin-ui] Version info


herflis:
- [x] discuss the future of the assemblies and installed packages part
- [x] add description for all the components 
- [x] add a link to the component's the its github repository
- [x] add changelog to the admin-ui part

- [x] remove versioninfo menuitem from the drawer

PR #502, opened 2019-12-05
Versioninfo item has removed from drawer

pusztaienike:


pusztaienike:
it closes #425

herflis:
@B3zo0 I think it can be completely removed now, the useful part of this feature (the action that returns the version info) is still available, so it will be easy to recreate the view based on Miklós' and Joe's needs and maybe this view will not be needed here on the admin-ui anyway, because they will use the snoversn ui in their daily work.

PR #503, opened 2019-12-05
deselection in select all function has fixed

pusztaienike:


packages/sn-list-controls-react/src/ContentList/ContentList.tsx
```diff
@@ -104,7 +104,7 @@ export interface ContentListProps<T extends GenericContent = GenericContent> {
   /**
    * Optional custom selection component
    */
-  getSelectionControl?: (selected: boolean, content: T) => JSX.Element
+  getSelectionControl?: (selected: boolean, content: T, callBack?: Function) => JSX.Element
```
B3zo0: I would not make this optional. This callback is needed for the custom component to work properly. Also the name should be onChange or onChangeCallback and its type is `() => void`

apps/sensenet/src/components/SelectionControl.tsx
```diff
@@ -3,12 +3,17 @@ import { GenericContent } from '@sensenet/default-content-types'
 import React from 'react'
 import { Icon } from './Icon'
 
-export const SelectionControl: React.FunctionComponent<{ isSelected: boolean; content: GenericContent }> = ({
-  isSelected,
-  content,
-}) => (
+export const SelectionControl: React.FunctionComponent<{
```
B3zo0: Please refactor the props to its own type.

PR #508, opened 2019-12-09
Feature/login revamp

B3zo0:
Changed the hash router to browser router.
Moved login from the desktop layout. It has its own page now.
login screen
Fixed the logout bug. (didn't redirect to login page)
Added [strict mode](https://reactjs.org/docs/strict-mode.html) to sn-app

Remaining tasks:

- [x] Add info to info icon

- Business Cat is an admin user with full access.
- Developer Dog is developer user who has control only over the content types, the settings and can see the version info.
- Editor Manatee is a content editor who can manage the non-system content items, build or save custom search queries.


apps/sensenet/src/components/login/info-box.tsx
```diff
@@ -0,0 +1,118 @@
...
+  createStyles({
+    infoBox: {
+      backgroundColor: theme.palette.primary.main,
+    },
+    marginBottom: {
+      marginBottom: theme.spacing(1),
+    },
+    listItemText: {
+      flex: '1 1',
+      textAlign: 'center',
```
herflis:
```suggestion
      textAlign: 'left',
```

PR #520, opened 2019-12-16
Feature/addbutton change

pusztaienike:
It is working what I expected, but we have 2 differences compare to the previous version:
- On **Content Types** root we have no action with the name of 'Add'
(https://dev.demo.sensenet.com/odata.svc/Root/System/Schema/ContentTypes/GenericContent/('ListItem')/Actions) so here Add button is not enabled  <==> In the previous version there was an Add button there
- On **Setup** root we have now Add button, cause we have 'Add' action here
(https://dev.demo.sensenet.com/odata.svc/Root/System/('Settings')/Actions) and we have so many child types what we can add: 
(https://dev.demo.sensenet.com/odata.svc/Root/System/('Settings')/EffectiveAllowedChildTypes?$select=Name&metadata=no) <==> In the previous version we did not have Add button on Setup


herflis:
@pusztaienike sorry, I forgot to mention that in case of ContentTypes the Action that should be available is not 'Add' but 'Upload'

pusztaienike:
> @pusztaienike sorry, I forgot to mention that in case of ContentTypes the Action that should be available is not 'Add' but 'Upload'

For Content type I had to track the queryData cause it has an allowedChildTypes attribute, what I needed for the Action list... 
Changes have been made. It is working now for Content types as well

apps/sensenet/src/hooks/use-drawer-items.tsx
```diff
@@ -60,6 +61,32 @@ export const useDrawerItems = () => {
     [localization],
   )
 
+  const getRootFromSetting = useCallback(
+    (item: ContentDrawerItem | QueryDrawerItem | BuiltinDrawerItem | DashboardDrawerItem) => {
+      switch (item.itemType) {
```
B3zo0: This should be added to the PersonalSettings.ts defaultSettings object.

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
```diff
@@ -63,28 +74,25 @@ const PermanentDrawer: React.FunctionComponent<RouteComponentProps> = props => {
...
-            ) : (
+            return (
               <NavLink
                 to={`/${btoa(repo.configuration.repositoryUrl)}${item.url}`}
                 activeStyle={{ opacity: 1 }}
                 style={{ textDecoration: 'none', opacity: 0.54 }}
-                key={index}>
-                <ListItem button={true}>
+                key={index}
+                isActive={match => {
```
B3zo0: I think this prop is not needed anymore.

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
```diff
@@ -121,11 +129,20 @@ const PermanentDrawer: React.FunctionComponent<RouteComponentProps> = props => {
                 />
                 <ListItemSecondaryAction>
                   {device === 'mobile' ? null : (
-                    <Link to={`/personalSettings`} style={{ textDecoration: 'none' }}>
+                    <NavLink
+                      to="/personalSettings"
+                      style={{ textDecoration: 'none' }}
+                      isActive={match => {
+                        if (!match) {
```
B3zo0: Please put the setCurrentPath logic to the onChange prop and remove this prop, also do this for the other navlink below.

PR #523, opened 2019-12-19
the text of allowchildtypes has been fixed

pusztaienike:
The style has also fixed, cause the text of a child type was unreadable in some cases

image

- [x] If the returned type's displayname and the given contenttypename is not equal hide the type from the add dropdownlist.
- [x] Check allowed childtypes lists on folders and their parents
**--> I found only the Setting menu where AllowedChildTypes is not set in the drawer item list
https://dev.demo.sensenet.com/OData.svc/Root/System/('Settings')/GetAllowedChildTypesFromCTD?metadata=no&$select=Name**


packages/sn-client-core/src/Schemas/SchemaStore.ts
```diff
@@ -57,7 +57,9 @@ export class SchemaStore {
     }
     let schema = this.schemas.find(s => s.ContentTypeName === contentTypeName) as Schema
     if (!schema) {
-      return this.getSchema(GenericContent)
+      const genericContent = this.getSchema(GenericContent)
+      genericContent.DisplayName = contentTypeName
```
B3zo0: While I see why you added this change I'm not sure if this is the right way. Lets discuss this on Monday

PR #524, opened 2019-12-19
client-core exports have been fixed

pusztaienike:


packages/sn-client-core/src/index.ts
```diff
@@ -1,6 +1,8 @@
 export { BypassAuthentication } from './Authentication/BypassAuthentication'
```
B3zo0: Please add and index.ts file in every sub directory and create a barrel file like this.
```js
export * from './Authentication'
export * from './Models'
export * from './Repository'
export * from './Schemas'
```

PR #525, opened 2019-12-22
Feature/theme switcher

pusztaienike:
design
- [x] Remove repository change from the usermenu
- [x] Refactor to avoid/remove theme service


B3zo0:
There is strange behavior when I change theme the drawer items gets reinitialized and there are multiple getActions request sent. Please check if you can avoid that.

apps/sensenet/src/context/ThemeProvider.tsx
```diff
@@ -14,11 +16,19 @@ const mergeThemes = (options: ThemeOptions, type: 'light' | 'dark') =>
   })
 
 export const ThemeProvider: React.FunctionComponent<{ theme: ThemeOptions }> = props => {
-  const ps = useContext(ResponsivePersonalSetttings)
-  const theme = mergeThemes(props.theme, ps.theme)
+  const injector = useInjector()
+  const service = injector.getInstance(PersonalSettings)
+  const settings = service.userValue.getValue()
+  const [pageTheme, setPageTheme] = useState<Theme>(mergeThemes(props.theme, settings.theme))
```
B3zo0: You should use personalSettings here instead of the service.userValue so we can avoid the useInjector, injector.getInstance... etc calls.

pusztaienike:
> There is strange behavior when I change theme the drawer items gets reinitialized and there are multiple getActions request sent. Please check if you can avoid that.

In use-drawer-item we are watching the **settings** (line 168-189) variable
`const settings = useContext(ResponsivePersonalSetttings)`
when we get the actions for drawer items.

`export const ResponsivePersonalSetttings = React.createContext(defaultSettings.default)`
... 
`ResponsivePersonalSetttings.Provider value={deepMerge(personalSettings.default, personalSettings.desktop)}>`

In ResponsiveContextProvider the **personalSettings** is reactive so when it changes it will effect that  we should update the drawer items as well.

To avoid this, we can remove the settings variable from the watches list in use-drawer-items, but it will cause lint warning...

pusztaienike:
> One more thing I would love to see in this pr is
> 
> ```js
> const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)')
> ```
> 
> this code somewhere.
> https://web.dev/prefers-color-scheme/
> It would be awesome if we would support this bydesign. 😊
> @herflis Are you okay with this?

**In the original issue light theme was what I should use:**

> Another related issue is that maybe we could try out 'light' as the default theme, because not everyone was fascinated by the dark theme. 🕶

PR #526, opened 2019-12-23
drawer revisit

pusztaienike:
- [x] Put the menu items' description to a tooltip


pusztaienike:
This will be complete with #520 and #525

herflis:
@pusztaienike Can we solve the following somehow?
If the drawer is closed the position of the tooltip is ok
image
But when the drawer is open the toolbar is way too off
image

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
```diff
@@ -1,43 +1,77 @@
...
-import { UserAvatar } from '../UserAvatar'
+import { useRepository } from '@sensenet/hooks-react'
+import clsx from 'clsx'
+import { createStyles, makeStyles, Theme } from '@material-ui/core'
+import { ResponsivePersonalSetttings } from '../../context'
+import { useDrawerItems, useLocalization, useSelectionService } from '../../hooks'
 import { AddButton } from '../AddButton'
 
+const useStyles = makeStyles((theme: Theme) => {
+  const primary = theme.palette.primary.main
```
B3zo0: There are unused styles here. navLinkActiveStyle, listItemIconDark, listItemIconLight.
The primary variable is not needed. You can inline the value. Also because of the unused styles the implementation does not match the design.

pusztaienike:
I have tried a new position for the tooltips.

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
B3zo0: I see hover is now working for the collapse icon and for the add button but not for other elements. Please add that styling to those elements as well.

PR #527, opened 2019-12-23
header revisit

pusztaienike:
I think we don't need the repository change here....it is duplicated

design
- [x] Put back to the repository dropdown
 --> **this pull request is about removing the repositoryselector - > Reject this PR to ignore it**


pusztaienike:
I closed this issue cause I only removed the repositoryselector here, what we need.

PR #528, opened 2019-12-30
new dropdown for addbutton

pusztaienike:
It will be complete with #520


apps/sensenet/src/components/AddButton.tsx
```diff
@@ -169,36 +184,28 @@ export const AddButton: React.FunctionComponent<AddButtonProps> = props => {
...
-              </Button>
-            </Tooltip>
+            <ListItem
+              key={childType.ContentTypeName}
+              button={true}
+              onClick={() => {
+                setShowSelectType(false)
+                openDialog({ name: 'add', props: { schema: childType, parent } })
+              }}>
+              <ListItemIcon>
```
B3zo0: I think it would make sense to wrap the icon and text with a tooltip element.

apps/sensenet/src/components/AddButton.tsx
```diff
@@ -139,28 +145,37 @@ export const AddButton: React.FunctionComponent<AddButtonProps> = props => {
...
           <ListItemText primary={localization.addNew} />
         </ListItem>
       )}
-      <SwipeableDrawer
-        anchor="bottom"
-        onClose={() => setShowSelectType(false)}
-        onOpen={() => {
-          /** */
+      <Popover
+        style={{ maxHeight: '85%' }}
```
B3zo0: We should set a min and max width for this element. @CsabaPeter please help us out here.

PR #529, opened 2019-12-31
useEffect has been modified

pusztaienike:
There are multiple actions call for one item and too many console error.

Try to refactor this code:

- [x]  get rid of console errors
--> there was no console error regarding use-drawer-items
- [X]  don't create duplicate requests
--> setDrawerItems and the checks are called now only once when the component is mounted... 
- [x]  try to minimize the request count
--> Now we call repo.getActions({ idOrPath: permission.path }) 6 times... but we need this, cause we have to check the actions of all the drawer items (except Search)


apps/sensenet/src/hooks/use-drawer-items.tsx
```diff
@@ -186,7 +186,8 @@ export const useDrawerItems = () => {
         return true
       })
       .then(items => setDrawerItems(items.map(item => getItemFromSettings(item))))
-  }, [getItemFromSettings, repo, repo.security, session, settings])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
B3zo0: This is not something I'm looking for. In local I don't see the duplicate requests in the develop branch right now. So this is probably safe to revert back.
There are console errors when the getActions request returns with 500 Access denied. Try to log in with editormanatee.
We should catch those errors and log it with the logger.

B3zo0: I'm okay with the request count :)

pusztaienike: ❔  Done.  But I think it is not a "user-friendly" solution. 
Now if you log in with editormanatee you will get 3 errors of 'Acces denied'... but you don't know why ..

B3zo0: I should have been more clear... This should be logged as debug not as an error.

PR #538, opened 2020-01-07
Fix the 'add new content' dialog

herflis:
- fixes misleading error message on unsuccessful save
- changes NewView's `parent` input param's type from GenericContent to string, because it was unnecessary to send a content when only its path is used


apps/sensenet/src/components/dialogs/add.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { Schema } from '@sensenet/default-content-types'
 import React from 'react'
 import { useLogger, useRepository } from '@sensenet/hooks-react'
 import { useLocalization } from '../../hooks'
 import { useDialog } from './dialog-provider'
 
 export type AddDialogProps = {
   schema: Schema
-  parent: GenericContent
+  parent: string
```
B3zo0: I think it would make sense to rename this property to parentPath then.

apps/sensenet/src/components/dialogs/add.tsx
```diff
@@ -26,7 +26,7 @@ export const AddDialog: React.FunctionComponent<AddDialogProps> = ({ schema, par
           handleCancel={closeLastDialog}
           repository={repo}
           contentTypeName={schema.ContentTypeName}
-          path={parent.Path}
+          path={parent}
```
B3zo0: While this is not part of the pr, passing down the path to NewView only to have it back on onSubmit feels wrong IMO.

apps/sensenet/src/components/dialogs/add.tsx
herflis: done

apps/sensenet/src/components/dialogs/add.tsx
herflis: this was necessary because of dms and the redux actions, but now I removed it from here and add the path at the related submit in the dms

apps/sensenet/src/components/AddButton.tsx
```diff
@@ -212,7 +212,7 @@ export const AddButton: React.FunctionComponent<AddButtonProps> = props => {
                 style={{ padding: '10px 0 10px 10px' }}
                 onClick={() => {
                   setShowSelectType(false)
-                  openDialog({ name: 'add', props: { schema: childType, parent } })
+                  openDialog({ name: 'add', props: { schema: childType, parentPath: props.path } })
```
B3zo0: Shouldn't this be props.parent.Path? When I select the it workspace under content and try to create a folder under it now it creates under /Content

herflis: When there was no selected element, props.parent was undefined so I changed the value of this prop now to a statement to handle both situations (with or without actual parent).

PR #549, opened 2020-01-14
Search FAB has been moved

pusztaienike:
We had to move the Search FAB from bottom-right cause of the chat app

### DIFF
--- a/apps/sensenet/src/components/SearchButton.tsx
+++ b/apps/sensenet/src/components/SearchButton.tsx
@@ -1,0 +1,121 @@
+import React, { useEffect, useState } from 'react'
+import { useRepository } from '@sensenet/hooks-react'
+import {
+  createStyles,
+  IconButton,
+  ListItem,
+  ListItemIcon,
+  ListItemText,
+  makeStyles,
+  Theme,
+  Tooltip,
+} from '@material-ui/core'
+import Add from '@material-ui/icons/Add'
+import { Link } from 'react-router-dom'
+import { useLocalization } from '../hooks'
+import { encodeQueryData } from './search'
+
+const useStyles = makeStyles((theme: Theme) => {
+  return createStyles({
+    mainDiv: {
+      display: 'flex',
+      justifyContent: 'center',
+      position: 'relative',
+    },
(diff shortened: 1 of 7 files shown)


apps/sensenet/src/localization/default.ts
```diff
@@ -338,7 +339,7 @@ const values = {
     openInSearchTitle: 'Open in Search',
     openInSearchDescription: 'Opens the query expression in the Search view',
     saveQuery: 'Save Query',
-    newSearch: 'New search',
+
```
B3zo0:
```suggestion
```

apps/sensenet/src/components/SearchButton.tsx
```diff
@@ -0,0 +1,121 @@
+import React, { useEffect, useState } from 'react'
+import { useRepository } from '@sensenet/hooks-react'
```
B3zo0: Please follow the naming convention for files. We use kebab-case for new files now.

apps/sensenet/src/components/SearchButton.tsx
```diff
@@ -0,0 +1,121 @@
...
+      },
+    },
+    navLinkListItem: {
+      width: '100%',
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'space-evenly',
+      textDecoration: 'none',
+    },
+    listDropdown: {
```
B3zo0: This style and the next one is not used. Please remove these.

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
```diff
@@ -108,7 +109,12 @@ const PermanentDrawer: React.FunctionComponent<RouteComponentProps> = props => {
             ) : null}
```
B3zo0: Can you also please get rid of the warning `Material-UI: the 'anchorEl' prop provided to the component is invalid.` by splitting the open and close button to have a separate Tooltip parent?

pusztaienike: This solution won't solve this warning :(

PR #558, opened 2020-01-17
drawer item activestyle has been modified

pusztaienike:
I had to create a new useEffect() in Permanent drawer, which watches the browser URL... if it changes I decode the URL into parts, and set the actual drawer item to a state from that info....
Not a "clean" solution, but we don't know for example the selectionService.activeContent after a relogin (logout/login)... the only permanent info was the location/URL.


B3zo0:
I think we should do the routing refactor first. This bug is low priority IMO and the logic here will likely fail after the routing refactor.

herflis:
Since the routing will be refactored, and the solution above will not be match the future solution, I close this pr and move the related issue back to the backlog. Issue should be revisited after the routing refactor.

PR #565, opened 2020-01-29
Feature/datagrid performance improve

pusztaienike:
- [x] Let's see if we can reduce the amount of the loaded content metadata and the number of requests here **-> I think we cannot reduce them**
- [x] check unnecessary renders in grid when actions menu opens [with](https://github.com/welldone-software/why-did-you-render) 
  **--> Avoid rerenderer the Permanent Drawer every time if selection changes**
**--> In some cases I used React.usereducer or React.useMemo**
**--> Moved the unnamed functions out from the props**
**--> Load settings has been refactored**
- [x] check actions request happens two times
**--> Tried to reduce requests: 
In AddButton we got the actions twice, now we get it only once - due to currentComponent watcher removal in useEffect() hook
(one request is needed cause the < actual content >.Actions is for other purpose**

+1:
- [x] Action menu opens only when you click the "..." on the active content (click twice)
**-->This bug has been fixed**
 
- [x] **Virtualized table component has been implemented -->and used 'everywhere' in snapp**
Virtualized Table is similar to Content List, but helps with performance issues if you have lots of items.
You can use it like the Content List component except onItemTap and onItemContextMenu props. These are not implemented in virtualized table.
It has some other differencies as well:

> <b><u>onRowDoubleClick</u></b>
> 
> It should be in a tableProps prop
> 
> ```tsx
> tableProps={{
>   onRowDoubleClick: //what to do on double click,
> }}
> ```
> 
> <b><u>fieldComponent</u></b>
> 
> I has been renamed to `cellRenderer`
> 
> <b><u>onRequestActiveItemChange</u></b>
> 
> This prop is not available in the Virtualized Table component, but you can define what should happen in the `onRowClick`
> 
> ```tsx
> onRowClick: rowMouseEventHandlerParams => {
>   //what to do
> }
> ```
> 
> <b><u>onItemClick</u></b>
> 
> This prop is not available in the Virtualized Table component, but you can define what should happen in the `onRowClick` as in the previous example
> 
> <b><u>onAction</u></b>
> 
> Not available in Virtualized Table component
> 
> <b><u>hideHeader</u></b>
> 
> This prop has been renamed to `disableHeader`, you can add it in the tableProps prop
> 
> ```tsx
> tableProps={{
>   disableHeader: false,
> }}
> ```
> For tableProps rowCount,rowHeight, headerHeight and rowGetter attributes are mandatory
> 
> Two features are missing from virtualized table what was exist in ContentList component:
> -----------------------------------------------------------------------------------------------------------
>   /**
>    * Called when a content item is tapped
>    */
>   onItemTap?: (e: React.TouchEvent, content: T) => void
> -----------------------------------------------------------------------------------------------------------  
>   /**
>    * Called when a user hits right click on a content item
>    */
>   onItemContextMenu?: (e: React.MouseEvent, content: T) => void
> -----------------------------------------------------------------------------------------------------------
>


packages/sn-list-controls-react/package.json
```diff
@@ -47,6 +48,7 @@
     "@types/react-dom": "^16.9.5",
     "@types/react-router": "^5.1.2",
     "@types/react-router-dom": "^5.1.2",
+    "@types/react-virtualized": "^9.21.7",
     "@types/uuid": "^3.4.7",
     "autoprefixer": "^9.7.1",
```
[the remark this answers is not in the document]
pusztaienike: import {
  AutoSizer,
  Column,
  Index,
  RowMouseEventHandlerParams,
  Table,
  TableCellProps,
  TableCellRenderer,
} from 'react-virtualized'  in virtualized-table.tsx

apps/sensenet/src/components/content-list/react-virtualized-table.tsx
```diff
@@ -100,9 +102,8 @@ export const CollectionComponent: React.FunctionComponent<CollectionComponentPro
       expand: ['CheckedOutTo', ...fields.filter(fieldName => isReferenceField(fieldName, repo))],
       orderby: [[currentOrder as any, currentDirection as any]],
     })
-    // loadSettings can NOT be added :(
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
pusztaienike: Unfortunatelly, not. We don't want to watch the "props" cause it will effect unnecessary rerenders.

apps/sensenet/src/components/content-list/react-virtualized-table.tsx
```diff
@@ -112,6 +113,10 @@ export const CollectionComponent: React.FunctionComponent<CollectionComponentPro
     setIsContextMenuOpened(false)
   }, [children, activeContent, selected])
```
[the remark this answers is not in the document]
pusztaienike: I think we don't need it. Removed!

packages/sn-list-controls-react/src/ContentList/virtualized-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+    selected: {
+      backgroundColor: theme.palette.type === 'dark' ? 'rgba(255, 255, 255, 0.08)' : 'rgba(0, 0, 0, 0.04)',
+    },
+    selectedLight: {
+      backgroundColor: 'rgba(255, 255, 255, 0.04)',
+    },
+  }),
+)
+
+interface Row {
```
B3zo0: This interface is the same as `Index` from react-virtualized

packages/sn-list-controls-react/src/ContentList/virtualized-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+}
+
+interface VirtualizedTableProps<T = GenericContent> extends ContentListBaseProps {
+  /**
+   * Contains custom cell template components
+   */
+  cellRenderer?: (props: VirtualCellProps) => React.ReactNode
+  /**
+   * Array of fields that should be displayed
+   * @default []
```
B3zo0: Where do you set this to empty array? Also if it has a default value then it should not be mandatory.

packages/sn-list-controls-react/src/ContentList/virtualized-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+   */
+  fieldsToDisplay: Array<keyof T>
+  tableProps: {
+    /** Number of rows in table. */
```
B3zo0: Please write single line comments in code like this
```javascript
// this function does nothing
const a = someFunction()
```
and in interfaces and types like this
```typescript
interface MyInterface {
  /**
   * This prop is useless
   */
   myUselessProp: string
}
```

packages/sn-list-controls-react/src/ContentList/virtualized-table.tsx
```diff
@@ -0,0 +1,305 @@
...
+    tableCell: {
+      flex: 1,
+    },
+    noClick: {
+      cursor: 'initial',
+    },
+    selected: {
+      backgroundColor: theme.palette.type === 'dark' ? 'rgba(255, 255, 255, 0.08)' : 'rgba(0, 0, 0, 0.04)',
+    },
+    selectedLight: {
```
B3zo0: Is this still needed?

PR #573, opened 2020-02-10
sn-client and sn-redux update with missing actions, odata params, etc

herflis:
As the new documentation is growing, some of the examples point to missing stuff in the API. This pr adds the following actions, params and related tests:

sn-client:
- `enableautofilters` OData param
- `enablelifespanfilter` OData param

sn-redux
- `getChildrenCount` action to use ${path}/$count
- `getProperty` action to use ${path}/${fieldname}
- `getPropertyValue` action to use ${path}/${fieldname}/$value
- `getMetadata` action to user ${path}/$metadata


packages/sn-redux/src/Actions.ts
```diff
@@ -519,3 +520,54 @@ export const setDefaultOdataOptions = (options: ODataParams<GenericContent>) =>
...
+export const getProperty = (idOrPath: string | number, propertyName: string) => {
+  const path = PathHelper.getContentUrl(idOrPath)
+  return {
+    type: 'GET_PROPERTY',
+    payload: (repository: Repository) => repository.fetch(`${path}/${propertyName}`),
+  }
+}
+
+/**
+ * Action creator for requesting a content's from sensenet Content Repository to get its value.
```
B3zo0:
```suggestion
 * Action creator for requesting a content's property from sensenet Content Repository to get its value.
```

packages/sn-redux/src/Actions.ts
```diff
@@ -122,6 +122,7 @@ import {
 } from '@sensenet/client-core'
 import { GenericContent, User } from '@sensenet/default-content-types'
 import { PromiseMiddlewareAction } from '@sensenet/redux-promise-middleware'
+import { PathHelper } from '@sensenet/client-utils'
```
B3zo0: Please add client-utils to package.json and to tsconfig.json as well.

herflis: I've added it to the package.json and I noticed that it was already there in the tsconfig's references array.  Is it okay there, or it should be added to one of the other lists there (e.g. includes)?

B3zo0: I can't see sn-client-utils inside the references. It should be there.

PR #574, opened 2020-02-10
Show upload only if it is possible to upload

pusztaienike:
The final style will be solved during the global style refactor.


herflis:
I don't know if it is related or not but after login the addbutton is active and after clicking it, it opens an empty menu

image

apps/sensenet/src/components/AddButton.tsx
```diff
@@ -173,20 +178,45 @@ export const AddButton: React.FunctionComponent<AddButtonProps> = props => {
...
+      })}>
       {!props.isOpened ? (
         <div className={classes.iconButtonWrapper}>
-          <Tooltip title={localization.addNew} placement="right">
+          {isAvailable ? (
+            <Tooltip
+              title={localization.addNew}
+              placement="right"
+              className={clsx({
+                [classes.disabled]: !isAvailable,
```
B3zo0: In `!isAvailable` state it renders another component so this class is not needed IMO.

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
```diff
@@ -67,11 +67,26 @@ const useStyles = makeStyles((theme: Theme) => {
   })
 })
 
+const isLocalization = (pathname: string) => {
+  const res = decodeURIComponent(pathname).split('/')
```
B3zo0: is this 2 function needed? can't we use the matchPath here as well?

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
```diff
@@ -99,7 +114,11 @@ const PermanentDrawer: React.FunctionComponent<RouteComponentProps> = props => {
           </div>
 
           {matchPath(props.location.pathname, `/:repositoryId/saved-queries`) === null ? (
-            <AddButton isOpened={opened} path={currentPath} />
+            matchPath(props.location.pathname, { path: `/`, exact: true }) === null &&
```
B3zo0: I would like to see this refactored. Lets talk about it sometime.

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
pusztaienike: Yes, we need them, cause in these cases we encode the Url datas (use-drawer-items.tsx line 83-133) We cannot get these info any other way. 
After the URL refactor, we should change this logic here as well (so it is just a temporary solution)

pusztaienike: The perfect solution was to create new componets for localization and users and groups with their own URL

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
pusztaienike: I changed the blacklist to a whitelist, so we don't need to create loop for the checks (cause it contains only 3 conditions)

apps/sensenet/src/components/MainRouter.tsx
```diff
@@ -22,6 +22,12 @@ const EditProperties = lazy(async () => await import(/* webpackChunkName: "editP
 const DocumentViewerComponent = lazy(async () => await import(/* webpackChunkName: "DocViewer" */ './DocViewer'))
 
 const TrashComponent = lazy(async () => await import(/* webpackChunkName: "Trash" */ './trash/Trash'))
+const UsersAndGroupsComponent = lazy(
+  async () => await import(/* webpackChunkName: "Trash" */ './users-and-groups/users-and-groups'),
```
B3zo0:
```suggestion
  async () => await import(/* webpackChunkName: "UserAndGroup" */ './users-and-groups/users-and-groups'),
```

apps/sensenet/src/components/MainRouter.tsx
```diff
@@ -22,6 +22,12 @@ const EditProperties = lazy(async () => await import(/* webpackChunkName: "editP
...
+)
+const LocalizationComponent = lazy(
+  async () => await import(/* webpackChunkName: "Trash" */ './localization/localization'),
```
B3zo0:
```suggestion
  async () => await import(/* webpackChunkName: "Localization" */ './localization/localization'),
```

PR #579, opened 2020-02-14
Make dms a 'simple' example

herflis:
- It is now available on https://sn-react-dms.netlify.com/ (dms.demo.sensenet.com will be unavailable)
- All the tracking codes and the privacy policy stuff are removed from the example
---
TODO:
- [x] remove all occurences of dmsservice
- list todos to make it work with dev.demo

- [x] change userprofile to one of the demo workspaces
- [x] fix usermenu (change DMSUserActions scenario to UserActions)
- [x] remove favorite ws feature
- [x] remove follow ws feature
- [x] fix ws list
- [x] registration action should be added to the dev.demo
- [ ] ~There're examples in the react-component-docs with login required. We should make a part of the dev.demo public to make these stuff available without login.~
- [x] remove user and group management pages and related features
- [x] change actionmenu scenario (from DMSListItems to ContextMenu)
- [x] breadcrumb contextmenu
- [x] document library celltemplates
- [x] batchaction scenario
- [x] add devdemo and demo users to the login screen
- [x] add sharing action to the client-core and sn-redux packages to make it more official
- [x] create a getSharing action to the sn-redux package and remove all the non-dms related stuff from the dms
- [x] shared with me page


examples/sn-react-component-docs/stories/pickers/ExampleApp.tsx
```diff
@@ -11,7 +11,7 @@ import {
 import React from 'react'
 
 const testRepository = new Repository({
-  repositoryUrl: 'https://dmsservice.demo.sensenet.com',
+  repositoryUrl: 'https://dev.demo.sensenet.com',
```
B3zo0: This is not going to work in dev.demo please the `currentPath` parameter in useListPicker.

examples/sn-dms-demo/src/store/shared.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (isFromAction(action, select)) {
+    return { ...state, selected: action.selected }
+  }
+  if (isFromAction(action, setActive)) {
+    return {
+      ...state,
+      active: action.active,
+    }
+  }
+  /** todo: auto-generated success action from @sn/redux */
```
B3zo0: Should this be done?

examples/sn-dms-demo/src/components/Shared.tsx
```diff
@@ -1,16 +1,185 @@
...
+            selected={this.props.selected}
+            onRequestSelectionChange={this.props.select}
+            active={this.props.active}
+            icons={icons}
+            items={this.props.sharedItems}
+            fieldsToDisplay={['DisplayName', 'ModifiedBy', 'ModificationDate']}
+            onRequestActiveItemChange={this.props.setActive}
+            displayRowCheckbox={false}
+            onRequestActionsMenu={(ev, content) => {
+              ev.preventDefault()
```
B3zo0: I see an error when I try to open context menu in shared view. I've logged in with devdog and businesscat shared a doc with me. `Uncaught TypeError: content.Actions.filter is not a function or its return value is not iterable`

packages/sn-redux/src/reducers/sharing.ts
```diff
@@ -0,0 +1,35 @@
+import { combineReducers, Reducer } from 'redux'
+import { SharingLevel, SharingMode } from '@sensenet/client-core'
+
+export interface SharingEntry {
```
B3zo0: This interface is the same as ODataSharingResponse from client-core. I think it would make sense to use that.

packages/sn-redux/src/Actions.ts
```diff
@@ -584,3 +585,52 @@ export const getSchema = () => ({
...
+ */
+export const share = (options: SharingOptions) => ({
+  type: 'SHARE',
+  payload: (repository: Repository) =>
+    repository.share({
+      content: options.content,
+      identity: options.identity,
+      sharingLevel: options.sharingLevel,
+      sharingMode: options.sharingMode,
+      sendNotification: options.sendNotification || true,
```
B3zo0: This is going to be true even if you don't want to send notification. I suggest to use the new nullish coalescing operator (??) in situations like this.
```suggestion
      sendNotification: options.sendNotification ?? true,
```

packages/sn-client-core/src/Repository/Repository.ts
```diff
@@ -301,6 +303,25 @@ export class Repository implements Disposable {
...
+    return await this.executeAction<{}, ODataSharingResponse>({
+      idOrPath: options.content.Id,
+      method: 'POST',
+      name: 'Share',
+      body: {
+        token: options.identity.toString(),
+        content: options.content,
+        level: options.sharingLevel,
+        mode: options.sharingMode,
+        sendNotification: options.sendNotification || true,
```
B3zo0: See my comment below.
```suggestion
        sendNotification: options.sendNotification ?? true,
```

examples/sn-dms-demo/src/store/shared.ts
herflis: It's not my comment, but I think it should be revisited later, when we return to finish the sharing dialog.

herflis:
@B3zo0 I've removed the related scenario from the profile action, so after the next daily rebuild there will be no 'My Profile' menuitem, Trash page is also added, however it is not possible to restore items from the trash (like on the trash page of the admin-ui), we should create an issue for that.

PR #591, opened 2020-03-10
fix frontend crash after deleting from tree

pusztaienike:
- [X] Current-content, selectionService.activeContent and selectionService.selection have been changed to the deleted item's parent.
- [X] Tests have been created for modify and delete action as well


herflis:
deleted content does not dissapear from the tree:

[screencast](https://gifyu.com/image/IGoL)

dialog buttons are not aligned well:

image

pusztaienike:
> deleted content does not dissapear from the tree:
> 
> [screencast](https://gifyu.com/image/IGoL)
> 
> dialog buttons are not aligned well:
> 
> image

**Deleting on root has ben fixed
Style will be fixed in https://github.com/SenseNet/sn-client/pull/593**

PR #617, opened 2020-04-02
DocViewer fixes

herflis:
- [x] toolbar is added, so now thumbnails and comments could be shown
- [x] additionally added to the toolbar: zooming, rotating and fitting functions
- [x] fix: adding comments returned error always, because the x and y coordinates were floating numbers and the request body was not a valid json. These two params' type is changed to string.

TODOs for the future (issues should be created):
(Each of the below feature has its placeholder, a button without the actual function implemented.)

- [ ] comments revamp and missing stuff
- [ ] download
- [ ] print
- [ ] save
- [ ] share
- [ ] watermark
- [ ] managing shapes

### DIFF
--- a/packages/sn-document-viewer-react/src/components/document-widgets/ToggleComments.tsx
+++ b/packages/sn-document-viewer-react/src/components/document-widgets/ToggleComments.tsx
@@ -11,10 +11,10 @@ export const ToggleCommentsWidget: React.FC = () => {
   const viewerState = useViewerState()
   return (
     <ToggleBase
-      isVisible={false}
+      isVisible={viewerState.showComments}
       title={localization.toggleComments}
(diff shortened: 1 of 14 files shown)


packages/sn-document-viewer-react/src/components/document-widgets/ToggleComments.tsx
```diff
@@ -11,10 +11,10 @@ export const ToggleCommentsWidget: React.FC = () => {
   const viewerState = useViewerState()
   return (
     <ToggleBase
-      isVisible={viewerState.showComments}
+      isVisible={false}
```
B3zo0, line 14: With this change we lose the ability to close the comments section. Is that something that you want?

PR #633, opened 2020-04-09
Download action to the contextmenu

herflis:
Download action was missing from the context menu, because Browse action is now used on Files as metadata browse as on other types of content (in the old sensenet, the Browse action on Files was actually download).
I've added a statement to check if the Browse action is available and the content's type is file and then put the action Download to the context menu. The underlying logic, the actual download method was already there in the js client code.

### DIFF
--- a/apps/sensenet/src/components/context-menu/content-context-menu.tsx
+++ b/apps/sensenet/src/components/context-menu/content-context-menu.tsx
@@ -44,6 +44,13 @@ export const ContentContextMenu: React.FunctionComponent<ContentContextMenuProps
       }
       const contentActions = contentFromCallback.Actions.filter((action) => !action.Forbidden)
 
+      if (contentActions.findIndex((action) => action.Name === 'Browse' && props.content.IsFile)) {
+        contentActions.push({
+          Name: 'Download',
+          DisplayName: 'Download',
+        } as ActionModel)
+      }
+
       if (isWriteAvailable(contentFromCallback)) {
         // If write is available it means that we have two actions. We want to show only the open edit for the user.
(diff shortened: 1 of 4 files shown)


apps/sensenet/src/components/context-menu/content-context-menu.tsx
```diff
@@ -44,6 +44,13 @@ export const ContentContextMenu: React.FunctionComponent<ContentContextMenuProps
       }
       const contentActions = contentFromCallback.Actions.filter((action) => !action.Forbidden)
 
+      if (contentActions.findIndex((action) => action.Name === 'Browse' && props.content.IsFile)) {
```
B3zo0, line 47: I would rather use `array.some` here. findIndex can return with 0 which is falsy but a valid index number.

packages/sn-default-content-types/src/DefaultContentTypes.ts
```diff
@@ -94,6 +94,8 @@ export class GenericContent {
   public Depth?: number
   /* This field is true if content is in a system folder/trash or the content is a system folder/file. */
   public IsSystemContent?: boolean
+  /* This field is true if content the content is a file. */
```
B3zo0, line 97:
```suggestion
  /* This field is true if the content is a file. */
```

apps/sensenet/src/components/context-menu/content-context-menu.tsx
herflis: Changed it `some` and changing it also highlighted another bug with the props.content so it is also changed to contentFromCallback :)

PR #568, opened 2020-01-31
authentication rework

B3zo0:
### client-core

- Changed RepositoryConfiguration from class to interface for better typing experience. It is not tightly coupled with GenericContent now.
  - ❗ sessionLifetime property is moved from repository configuration to JWTService
  - 🆕 **token** can be passed to the repository configuration. This will be sent with all requests as an Authorization header.

### default-content-types

- 🆕 AllFieldNames type! This can be used in places where we used `Array<keyof GenericContent>`. This type extracts all the keys of fields from every content type we have in default content types.

### snapp

- useContentRouting hook is refactored to a simple class that needs repository object to instantiate. Also it is renamed to ContentContextService.
- Repository selector is removed.
- repositories and lastRepository props are removed from personal settings.
- useSession is not valid anymore. With oidc auth we don't store any information on the reposiotry object. This is replaced with useCurrentUser which holds the value of the current sensenet user. You can get the oidc user from useOidcAuthentication hook.

## Todo

- [x] Convert an example app to use the new authentication method
- [x] Create package for oidc-authentication components
- [x] ~Move `GetClientRequestParameters` action to repository~ `// This is not needed because this action is only needed for admin-ui`
- [x] Include min version of oidc-client on production build // This is included by default
- [x] Add tests to `@sensenet/authentication-oidc-react` package
- [x] Change all example app to use the new authentication
  - [x] ~Create a new package with common components (login, repositoryProvider with auth)~ `// not going to do this in this pr`
- [x] Create documentation ~& blog post about the changes, a tutorial maybe?~ `// basic documentation is added to the readme.md of @sensenet/authentication-oidc-react`

PR #634, opened 2020-04-14
Change landing page after login

pusztaienike:


apps/sensenet/src/components/login/Login.tsx
```diff
@@ -112,7 +111,7 @@ export const Login = () => {
         })
         if (match.path === '/login') {
           await sleepAsync(1800)
-          const { from } = location.state || { from: { pathname: `/${btoa(repoToLogin.configuration.repositoryUrl)}` } }
+          const { from } = { from: { pathname: `/${btoa(repoToLogin.configuration.repositoryUrl)}` } }
           history.replace(from)
```
B3zo0, line 115: While this is going to be removed in #568 . We can merge this if you want. The desired behavior should be also in the new auth pr. 
```suggestion
          history.replace(`/${btoa(repoToLogin.configuration.repositoryUrl)}`)
```

PR #639, opened 2020-04-15
Create loader for move/copy

pusztaienike:
-  [x] create loader for move/copy 
To keep the "unity" in admin-ui I have created the same logic as in delete dialog.


B3zo0:
I think it would make sense to let the user cancel the copy-move dialog if he/she wants.
Right now you can click outside the dialog and it will close but not cancel the request.
1. cancel button should not be in disabled state
2. the request should be cancelable
3. we should let or not let the user exit the dialog by clicking outside. If we allow it it should cancel the request. if we not allow it then the only way a user can exit the dialog is the cancel button.

pusztaienike:
> I think it would make sense to let the user cancel the copy-move dialog if he/she wants.
> Right now you can click outside the dialog and it will close but not cancel the request.
> 
> 1. cancel button should not be in disabled state
> 2. the request should be cancelable
> 3. we should let or not let the user exit the dialog by clicking outside. If we allow it it should cancel the request. if we not allow it then the only way a user can exit the dialog is the cancel button.

I have disabled the close on "click away" and "esc key" actions on these 3 batch action dialogs.
The user should not cancel the requst because it can cause some not intended behaviour.

PR #646, opened 2020-04-21
Fix default value of rootPath on explore component

taki9:
Explore component rendered empty because rootPath was undefined.

- [x] Container item page render correctly


B3zo0:
I think it make sense to remove this condition from explore then.
https://github.com/SenseNet/sn-client/blob/develop/apps/sensenet/src/components/content/Explore.tsx#L72

PR #658, opened 2020-04-30
Docviewer stlye revamp

pusztaienike:
- [X] When opening a document the command palette's icon in the header becomes black
- [X] The docviewer's header should be the same color as the sidebar with no shadow
- [X] The icons color should be the same as the sidebar icons, when active the color should be #26A69A
- [X] The comment icon is too far away from the close button
- [X] When the comments are opened the add comment link should be under the title
- [X] The cancel button should be outlined Screen Shot 2020-04-23 at 14.48.14.png -> should be white/black without outlined style

- [X] Question: Is there a need for multiple rotate functions (there are two separate right now) and multiple resizing tools?
Yes. One is for the page and one is for the document

+1: Approve dialog button style change
image


packages/sn-document-viewer-react/src/components/document-widgets/ToggleThumbnails.tsx
```diff
@@ -1,12 +1,12 @@
-import React from 'react'
-import { styles } from '../styles'
+import React, { CSSProperties } from 'react'
+//import { styles } from '../styles'
```
zoltanbedi, line 2:
```suggestion
```

packages/sn-document-viewer-react/src/components/document-widgets/ToggleThumbnails.tsx
```diff
@@ -17,8 +17,11 @@ export const ToggleThumbnailsWidget: React.FC = () => {
       <svg
         version="1.1"
         xmlns="http://www.w3.org/2000/svg"
-        style={{ width: '1em', height: '1em' }}
-        fill={viewerState.showThumbnails ? styles.colors.icon.inactive : styles.colors.icon.active}
+        style={
+          viewerState.showThumbnails
+            ? { width: '1em', height: '1em', fill: props.activeColor }
```
taki9, line 21: props.style not needed in this case?

PR #669, opened 2020-05-07
Layout standardization

pusztaienike:
### Main changes:
I have created a new service (called: DialogActionService) where we can store the actual action and we can use it from every components as a hook (useDialogActionService). It has only two observable values: activeAction and contentTypeNameForNewContent. The contentTypeNameForNewContent is needed only for 'new' action, cause we should pass the contentTypeName for the New component depending on which Type is selected from the AllowedChildTypes list under AddButton.
If no action is active the activeAction should be undefined.

I also removed every props which are unneccesary cause of the new version of action handling.
Previous solution was: [this](https://teams.microsoft.com/l/file/FE8D2FA8-6845-4C05-AD04-A8CD3F9ECF72?tenantId=6069be31-3d96-48fc-af43-2ed5a1538540&fileType=pdf&objectUrl=https%3A%2F%2Fsensenet.sharepoint.com%2Fsites%2Fsnteam%2FMegosztott%20dokumentumok%2FGeneral%2Fdiagrams%2Ffrontend%20related%20diagrams%2FPropsRelatedEditAndBrowseInExplore.pdf&baseUrl=https%3A%2F%2Fsensenet.sharepoint.com%2Fsites%2Fsnteam&serviceName=teams&threadId=19:1faf3f23c2424349b60b15aa7bb80569@thread.skype&groupId=923b6850-5d6d-4acb-a28c-f15ae7b409e2)

--------
### Corresponded issue:
The visibility of the tree and the breadcrumb is quite confusing now on the admin-ui and should be standardized a bit.

- [x] **edit properties** -> tree + breadcrumb
- [x] **new** -> tree + breadcrumb
- [x] **browse** -> tree + breadcrumb
- [x] **versions** -> tree + breadcrumb
- [x] **edit binary** -> filename and a back/close button that navigates back to the related list (content types, setup, localization)
- [x] **wopiedit/wopiview** -> filename and a back/close button
- [x] **doc preview** -> theres a close button on its own toolbar, but it would be better if it will be like at the other ones with a toolbar (e.g. edit binary)
- [x] **delete** -> it is good as is it is ✔

- [x] **users&groups** will have a custom ui but here could be also both tree and breadcrumb visible in the future, but here the breadcrumb root should be /Root/IMS/Public, the user should not be able to navigate above that
--------

### Action plan:
- [ ] **set permission**s 
--> tree + breadcrumb
- [ ] **if I open a CTD, Settings or Localization file to edit the related drawer item forgets that it is active**
--> It will be solved after refactor of urls and components
- [ ] **should use query builder instead of this**
apps/sensenet/src/components/content/Commander.tsx
`history.push(`${applicationPaths.newProperties}?path=${activeParent.Path}`)`
-->Commander will be removed in the future

- [ ] **dependency injection is not needed**
apps/sensenet/src/hooks/use-dialogaction-service.ts
-->It would be a big task to avoid the usage of injection -> should plan it later

- [ ] **onNavigate function is duplicated**
apps/sensenet/src/components/users-and-groups.tsx
--> onNavigate should be a hook, and should be refactored content/index.tsx as well 
[onNavigateSecondary can be removed after Commander will be deleted, onNavigate should be changed with the hook what we create later]
 
- [x] **eslint rule for ordering imports**
not related to this PR, New issue is opened for this #683


herflis:
Some related (and non-related) minor issues that I think should be fixed in the context of this task:

- [ ] on the edit form the icon in the title should be a link and when a user clicks it he should be navigated to the related CTD (if it is a bigger task that is seems, we could move it to a separate issue)
- [ ] in titles (on forms or dialogs) the actual content name part should be a bit more highlighted (change its font-weight to 500)
image
- [ ] modifiedby field on browse view is empty
image
- [ ] last column's title on the versions table rather should be simply 'Restore' 
- [ ] if I open a CTD, Settings or Localization file to edit the related drawer item forgets that it is active

pusztaienike:
### Findings from Zoli

- [ ] During "Add new" process Addbutton should be disabled
---
**apps/sensenet/src/components/MainRouter.tsx**
`const VersionProperties = lazy(() => import(/* webpackChunkName: "newProperties" */ './version/version-properties')) `
- [ ]  wrong chunkname
---
**apps/sensenet/src/components/content-list/content-list.tsx**
- [ ]  should avoid variables like 'a' or 'c' 
---
**apps/sensenet/src/components/content-types.tsx**
`// eslint-disable-next-line react-hooks/exhaustive-deps`
- [ ] only if it really needed, should avoid this otherwise
---
**apps/sensenet/src/components/content/Commander.tsx**
`history.push(`${applicationPaths.newProperties}?path=${activeParent.Path}`)`
- [ ] should use query builder instead of this
---
**apps/sensenet/src/components/content/Explore.tsx**
`<VersionView handleCancel -> `
- [ ] async is not needed 
`onActivateItem={(content) => onActivateItemOverride(content)} `
- [ ] should be onActivateItem={onActivateItemOverride}
---
**apps/sensenet/src/components/localization.tsx
apps/sensenet/src/components/search/saved-queries.tsx
apps/sensenet/src/components/setup/setup.tsx**
`// eslint-disable-next-line react-hooks/exhaustive-deps`
- [ ] duplicated parts of code
---
**apps/sensenet/src/components/new/new-properties.tsx**
- [ ]  handleCancel - async is not needed 
---
**apps/sensenet/src/components/tree/tree-with-data.tsx**
`!itemCount || itemCount == null`
- [ ] the first condition is stronger than the second (second is not needed)
---
**apps/sensenet/src/components/users-and-groups.tsx**
- [ ] onNavigate function is duplicated
---
**apps/sensenet/src/components/view-controls/edit-view.tsx**
- [ ] magic pixels should be stored in "global variables"
    here as well: 
        - apps/sensenet/src/components/view-controls/version-view.tsx és
        - apps/sensenet/src/components/view-controls/new-view.tsx
        - apps/sensenet/src/components/view-controls/version-view.tsx
---
**apps/sensenet/src/components/view-controls/new-view.tsx**
` props.submitCallback && props.submitCallback()`
- [ ] props.submitCallback?.()
---
**apps/sensenet/src/components/view-controls/version-view.tsx**
- [ ] not a form
---
**apps/sensenet/src/components/view-controls/version-view.tsx**
- [ ] import Button from '@material-ui/core/Button' andimport { } from '@material-ui/core' -> should use only one method
- [ ] if (content === undefined) return null - else is not needed, cause of return statement
- [ ] props.handleCancel && props.handleCancel() -> props.handleCancel?.()
---
**apps/sensenet/src/hooks/use-dialogaction-service.ts**
- [ ] dependency injection is not needed
---
**apps/sensenet/src/components/version/version-properties.tsx
apps/sensenet/src/components/edit/edit-properties.tsx
apps/sensenet/src/components/browse/browse-properties.tsx**
- [ ]     HOC?
---
- [ ] eslint rule for ordering imports

apps/sensenet/src/components/content/Explore.tsx
```diff
@@ -95,45 +121,94 @@ export function Explore({ currentPath, onActivateItem, onNavigate, rootPath, fie
...
+                            ) : (
+                              <span>
+                                Info about{' '}
+                                <span style={{ fontWeight: 500 }}>
+                                  {selectionService.activeContent.getValue()?.DisplayName}
+                                </span>
+                              </span>
+                            )}
+                            <span
+                              onClick={async () => {
```
herflis, line 153:
```suggestion
title={`Open ${selectionService.activeContent.getValue()!.Type} CTD`}
                              onClick={async () => {
```

PR #703, opened 2020-06-11
Browse type cleanup

taki9:


herflis:
It's probably not related to this issue, but in the trash there's a myterious padding on the left

image

PR #707, opened 2020-06-17
Protected sn routes & custom content pages

taki9:


apps/sensenet/src/hooks/use-drawer-items.tsx
```diff
@@ -44,26 +109,24 @@ export const useDrawerItems = () => {
...
         case 'ContentTypes':
           return <Widgets />
         case 'Localization':
           return <Language />
-        case 'Trash':
-          return <Delete />
         case 'Setup':
           return <Build />
+        case 'CustomContent':
+          return <Folder />
```
herflis, line 129:
```suggestion
          return item.settings && item.settings.icon ? (
            <Icon item={{ ContentTypeName: item.settings.icon }} />
          ) : (
            <Folder/>
          )
```

herflis:
There was a setting on drawer items where the users could set which columns should be displayed on the grid
{
  "itemType": "Content",
  "settings": {
    "description": "Explore your content",
    "icon": "Workspace",
    "title": "Content",
    **"columns": ["DisplayName", "ModifiedBy", "ModificationDate"],**
    "browseType": "explorer",
    "root": "/Root"
  }
}

It would be awesome to add/keep this functionality in the new solution if it is not a huge work 🙂

PR #790, opened 2020-08-13
Dashboard revamp

taki9:
- the price of the plan is missing
- what is the unit of storage space?


pusztaienike:
ERROR in src/components/dashboard/index.tsx:1:17
TS6133: 'User' is declared but its value is never read.
  > 1 | import { Group, User } from '@sensenet/default-content-types'

herflis (approved):

PR #800, opened 2020-08-24
Permission editor dialog

pusztaienike:
- [X]  Clicking on a user or group in the entry list a dialog should open
- [X]  The dialog should be positioned right to see the selected content while you are setting permissions
- [X]  The dialog has a header with the FullName of the user or with the DisplayName of the group
- [X]  The dialog also has a button row with a Cancel and a Save button. The Cancel button closes the dialog, and with the Save you can save the changes that you've made (that means that the changes are saved only when the Save button is clicked and not when one of the switches is clicked)
- [X]  The permissions are grouped into categories. These categories are listed in the left part of the dialog. List of permissions and permission 'groups' are stored in '/Root/System/Settings/Permission.settings'
- [X]  The right part of the dialog contains the actual permissions that belong to the selected group
- [X]  Each item in the permissions list has a switch so the permission settings can be changed one by one
- [X]  The switches have two states null and allow (deny is not an option for now)
- [X]  The permission groups have a switch as well for switching on or off all the related permissions. So if you check the Read switch all the permissions of the Read group should be switched on.
- [X]  There's an item in the list on the left for granting full access. Clicking this link switches all groups and individual permissions on or off.
- [X]  When the dialog opens, the switches should be switched on or off based on the settings in the chosen entry
- [X]  If you open the dialog for an inherited entry the allowed permission settings cannot be overridden and should be disabled but the untouched permissions can be edited freely


apps/sensenet/src/components/dialogs/permission-editor-dialog.tsx
```diff
@@ -0,0 +1,396 @@
...
+            />
+          </ListItem>
+          <Divider />
+          <ListItem>
+            <ListItemText primary={localization.permissionEditor.localOnly} className={classes.disabled} />
+            <Switcher checked={false} size="small" disabled onClick={() => {}} />
+          </ListItem>
+        </List>
+        <div className={clsx(classes.column, classes.rightColumn)}>
+          <div className={classes.permissionContainer}>
```
taki9, line 316: remove this div and move the class into List

apps/sensenet/src/components/dialogs/permission-editor-dialog.tsx
```diff
@@ -0,0 +1,396 @@
...
+              {localization.forms.cancel}
+            </Button>
+            <Button
+              aria-label={localization.forms.submit}
+              color="primary"
+              variant="contained"
+              autoFocus={true}
+              onClick={async () => {
+                if (props.path) {
+                  //TODO: rewrite this when it will has a return value (JSON)
```
taki9, line 370: it is done in a later PR, right?

apps/sensenet/src/components/dialogs/permission-editor-dialog.tsx
```diff
@@ -0,0 +1,396 @@
...
+              onClick={closeLastDialog}>
+              {localization.forms.cancel}
+            </Button>
+            <Button
+              aria-label={localization.forms.submit}
+              color="primary"
+              variant="contained"
+              autoFocus={true}
+              onClick={async () => {
+                if (props.path) {
```
taki9, line 369: should we give some feedback if no path given? or is that a valid scenario?

taki9 (changes requested):

taki9:
I could imagine a Reset button on the dialog which reset to the saved state.

apps/sensenet/src/components/dialogs/permission-editor-dialog.tsx
pusztaienike: honestly we don't need the response value, cause we cannot use it

apps/sensenet/src/components/dialogs/permission-editor-dialog.tsx
pusztaienike: You cannot open the permission view if there is no currentContent (because of context-menu - it has to reqest the actions so it also need a content), so it should not be undefined

taki9 (approved):

PR #804, opened 2020-08-25
Make lighthouse report better

pusztaienike:
Improve accessibility from 84% to 94%


herflis (changes requested):
Pls move this texts to the localization file, thanks 💙

herflis (approved):

PR #813, opened 2020-08-27
Cleanup settings

herflis:
- settings descriptions are moved to the sensenet repository as the Settings contents' descriptions
- WellknownContentCard are renamed to ContentCard and every wellknow... related stuff is removed because from now on the settings and their descriptions come from the contents themselves (no 'Other settings' list etc.)


pusztaienike (approved):
On netcore test DocumentPreview.settings and Portal.settings description are missing.
image

taki9 (approved):

PR #822, opened 2020-08-31
Fix reference field as column on custom menu

taki9:


herflis (approved):

pusztaienike (changes requested):
For me  it is not working properly. With the example of the related issue I got always the DisplayName of the "RelatedRepository" even if I set the Column to "RelatedRepository/DataSource", or "RelatedRepository/Host" etc.
image

pusztaienike (approved):

PR #834, opened 2020-09-02
Permission editor - Make content public

pusztaienike:
Add a button next to the add new entry button. The button text should be displayed by the actual state of the content (public/private). The GetAcl action should contain the needed information.

- [X]  Display the button based on the content's state in the GetAcl response
- [X]  Add a tooltip for the button, where we can add more info about the feature in-deep
- [X]  If the user clicks the button based on the current state a SetPermission action should be sent to the server with the identity /Root/IMS/BuiltIn/Portal/Visitor and with the property "Open": "allow"


herflis (approved):

apps/sensenet/src/components/view-controls/permission-view.tsx
```diff
@@ -62,9 +63,21 @@ const useStyles = makeStyles((theme: Theme) => {
...
+      backgroundColor: '#12cdd4',
+      color: theme.palette.common.white,
+      '&:hover': {
+        backgroundColor: '#00838f',
+      },
+    },
   })
 })
 
+const VISITOR_PATH = '/Root/IMS/BuiltIn/Portal/Visitor'
```
taki9, line 79: import { ConstantContent } from '@sensenet/client-core'
ConstantContent.VISITOR_USER.Path

taki9 (changes requested):

PR #835, opened 2020-09-02
Permission editor dialog additional features

pusztaienike:
- [X] CSS finalization

- [x] LocalOnly flag can be switched on or of for entries. The link is available in the left column of the dialog and has not got any content to open in the right column, so clicking on this link switches only its own switch and the localOnly flag in the background
- [x] For groups the user can open and manage group members as well
   - [x] In case of groups the 'Member' menuitem should be shown in the left column of the dialog and it should list group members in the right column when the user clicks it
   - [x] This member list can work as the same as the member manager list/dialog in the users&groups page, so adding or removing a member is requested to the server immediately, the dialogs save and cancel button are for the whole permission settings process not for membership changes


herflis (approved):

apps/sensenet/src/components/dialogs/permission-editor/permission-editor-members.tsx
```diff
@@ -0,0 +1,127 @@
+import { reactControlMapper } from '@sensenet/controls-react'
```
taki9, line 1: it is almost the same exact code as in reference-content-list, we should create a new common component, which is customizable with props for the unique requirements. or make reference-content-list customizable for changes in these case, if that possible

taki9: ok, now i see, you tried it, could we improve that?

taki9 (changes requested):

taki9 (approved):

PR #838, opened 2020-09-04
Permission editor - add new entry

pusztaienike:
- [X] Implement assign new entry feature
- [X] Fix selected CSS style on List Items when open permission-editor-dialog

User should be able to add new permission entries in the permission editor.
1. Clicking on the button a dialog should open with an autocomplete textbox, that queries users and groups from /Root/IMS/Public only. The autocomplete could be the same as the one in the users&groups 'add member' dialog.
2. If the user clicks add button on the dialog, the chosen user or group should be added as the last item of the entry list.
3. The newly added entry should be selected immediately with opening the permission editor dialog.


herflis (approved):

apps/sensenet/src/components/dialogs/member-select-dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+          actionName="new"
+          fieldOnChange={(_fieldName, newMember) => setSelectedMember(newMember)}
+          repository={repository}
+          settings={
+            {
+              AllowedTypes: ['User', 'Group'],
+              SelectionRoots: ['/Root/IMS/Public'],
+              Name: 'Name',
+              DisplayName: 'Name',
+              Description: 'Type in a name',
```
taki9, line 78: localization

apps/sensenet/src/components/dialogs/permission-editor-dialog.tsx
```diff
@@ -337,7 +337,10 @@ export function PermissionEditorDialog(props: PermissionEditorDialogProps) {
             <Button
               aria-label={localization.forms.cancel}
               className={globalClasses.cancelButton}
-              onClick={closeLastDialog}>
+              onClick={() => {
+                closeLastDialog()
+                props.callBackFunction?.()
```
taki9, line 342: it is not the best solution, is it possible to refesh right after new entry has added or here, but only in that case?

apps/sensenet/src/components/view-controls/permission-view.tsx
```diff
@@ -131,6 +132,25 @@ export const PermissionView: React.FC<PermissionViewProps> = (props) => {
     visitorEntries && visitorEntries.length > 0 ? setIsPrivate(false) : setIsPrivate(true)
   }, [permissions])
 
+  const selectMemberFunction = (newEntry: EntryType) => {
+    if (permissions) {
+      const localPermissions = { ...permissions }
```
taki9, line 137: permissions.entries.push(newEntry) do the same, the permission.entries handled by reference in both cases

apps/sensenet/src/components/view-controls/permission-view.tsx
```diff
@@ -131,6 +132,25 @@ export const PermissionView: React.FC<PermissionViewProps> = (props) => {
...
+      localPermissions.entries.push(newEntry)
+    }
+    setSelectedListItem(`${newEntry.identity.id}${newEntry.propagates}`)
+    openDialog({
+      name: 'permission-editor',
+      props: {
+        entry: newEntry,
+        path: currentContent!.Path,
+        callBackFunction: () => {
+          setRefreshFlag(!refreshFlag)
```
taki9, line 147: use two parameters insted: submitCallback, cancelCallback
submitCallback is the current callbackFunction
cancelCallback: () => permission.entries.pop()

apps/sensenet/src/components/view-controls/permission-view.tsx
```diff
@@ -266,10 +301,27 @@ export const PermissionView: React.FC<PermissionViewProps> = (props) => {
...
+              <List id="setOnThisList" component="div" disablePadding>
                 {setOnThisArray?.map((setOnThisEntry) => {
                   return (
-                    <ListItem button key={setOnThisEntry.identity.id}>
+                    <ListItem
+                      selected={selectedListItem === `${setOnThisEntry.identity.id}${setOnThisEntry.propagates}`}
+                      button
+                      key={`${setOnThisEntry.identity.id}${setOnThisEntry.propagates}`}
+                      onClick={() => {
+                        setSelectedListItem(`${setOnThisEntry.identity.id}${setOnThisEntry.propagates}`)
```
taki9, line 312: its format is interesting 😄

taki9 (changes requested):

taki9 (approved):

PR #839, opened 2020-09-04
Repository should not reassign after token refresh

taki9:


pusztaienike (commented):
I think it is still has a bug. If I open the businesscat user and change its name to businesscat1 without save the token renewal will update it to businesscat again.

pusztaienike (approved):
🦓

herflis (approved):

PR #850, opened 2020-09-08
Reference field cell pimp-up

herflis:
Reference field's 'cell template' is updated with making the value a link to the referenced content's edit view.

ref_field


pusztaienike (approved):
🐔

PR #883, opened 2020-09-11
Fix email field style

pusztaienike:


herflis (approved):

apps/sensenet/src/components/content-list/email-field.tsx
```diff
@@ -1,14 +1,34 @@
...
 
+const useStyles = makeStyles(() => {
+  return createStyles({
+    anchor: {
+      whiteSpace: 'nowrap',
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+    },
+    cell: {
+      padding: '0 3px',
```
taki9, line 14: no other field has left padding, i think right padding would be enough

taki9 (approved):

PR #885, opened 2020-09-11
Prevent multiple upload request on CTDs

pusztaienike:


herflis (approved):

apps/sensenet/src/components/editor/text-editor.tsx
```diff
@@ -46,6 +46,7 @@ export const TextEditor: React.FunctionComponent<TextEditorProps> = (props) => {
 
   const saveContent = async () => {
     try {
+      setHasChanges(false)
```
taki9, line 49: on catch case, you should set it back to true, not?

taki9: please fix this in new-file-text-editor too

taki9 (changes requested):

taki9 (approved):

PR #898, opened 2020-09-14
Fix active page number

pusztaienike:


herflis (approved):

packages/sn-document-viewer-react/src/models/viewer-state.ts
```diff
@@ -68,4 +68,9 @@ export interface ViewerState {
    *
    */
   onPageChange: ObservableValue<number>
+
+  /**
+   *
+   */
+  visiblePages: number[]
```
taki9, line 75: why visiblePages is needed at all? activePages is not good for this purpose?

taki9 (changes requested):

packages/sn-document-viewer-react/src/components/PageList.tsx
```diff
@@ -138,9 +138,11 @@ export const PageList: React.FC<PageListProps> = (props) => {
     }
 
     setMarginTop(_marginTop)
-    setMarginBottom(_marginBottom)
+    setMarginBottom(_marginBottom === 0 ? 20 : _marginBottom)
```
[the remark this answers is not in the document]
pusztaienike: marginBottom cannot be 5 I think
image

packages/sn-document-viewer-react/src/models/viewer-state.ts
pusztaienike: it will be an infinite loop :(

taki9 (approved):

PR #922, opened 2020-09-16
Fix reference link and addnew button on root screen

herflis:
- AddNew button was missing (at least for me in the snover repo), so I've added it to the list of app path, where the AddNew button is visible
- Reference link's color is fixed
- Reference field link is fixed for low level contents


pusztaienike (approved):
🦓

apps/sensenet/src/components/content-list/reference-field.tsx
```diff
@@ -21,6 +21,10 @@ const useStyles = makeStyles(() => {
       marginLeft: '1em',
       textTransform: 'lowercase',
     },
+    link: {
+      marginLeft: '1em',
```
taki9, line 25: marginRight should be on the icon, because this margin will be there even if no icon rendered

apps/sensenet/src/components/drawer/PermanentDrawer.tsx
```diff
@@ -129,6 +129,7 @@ export const PermanentDrawer = () => {
               PATHS.setup.appPath,
               PATHS.contentTypes.appPath,
               PATHS.localization.appPath,
+              PATHS.root.appPath,
```
taki9, line 132: PATHS.custom.appPath.replace(':path', 'root'),
and root can be deleted from PATHS

taki9 (changes requested):

taki9 (approved):

PR #992, opened 2020-10-01
Fix style of permission view buttons

pusztaienike:


apps/sensenet/src/components/view-controls/permission-view.tsx
```diff
@@ -42,9 +42,11 @@ const useStyles = makeStyles((theme: Theme) => {
     },
     title: {
       fontSize: '20px',
+      paddingRight: '10px',
     },
-    textBolder: {
+    contentName: {
       fontWeight: 500,
+      wordBreak: 'break-word',
```
taki9, line 49: it is working with the default value (normal) of wordBreak, so you can delete this

taki9 (changes requested):

apps/sensenet/src/components/view-controls/permission-view.tsx
pusztaienike: but it is not working with some display size 
I think it is better with break
permission_button

herflis (approved):

apps/sensenet/src/components/view-controls/permission-view.tsx
```diff
@@ -42,8 +44,14 @@ const useStyles = makeStyles((theme: Theme) => {
     },
     title: {
       fontSize: '20px',
+      paddingRight: '10px',
+      width: `calc(100% - ${permissionActionbuttonsWidth}px)`,
```
taki9, line 48: I think you could remove width and margin set, and get the same result without magic numbers.

taki9 (changes requested):

apps/sensenet/src/components/view-controls/permission-view.tsx
pusztaienike: without these it seems like the screenshot:
image

PR #1003, opened 2020-10-07
Handle UI refresh after content actions

taki9:
extra fixes not directly related to the connected issue:
- boolean input field was checked when default value is false
- bug fixed in getUrlForContent (in rare cases navigated to another menu with invalid path)
- edit view not displayed error after save
- view title icon


pusztaienike (approved):
👯‍♀️

herflis (approved):

Issue #978, opened 2020-09-29
[BUG] Edit after edit error


pusztaienike:
If I open an edit form when another edit form is open sometimes* it throws error

* if the edit form contain the same fields then not

## 📦 Package / Application

admin-ui

## 👉 Steps To Reproduce

1. Go to Content menu
2. Click on Sample Workspace in the tree
3. Right click on Sample Workspace
4. Select Edit from dropdown
5. Right click on Document Library in the tree (under Sample Ws)
6. Select Edit from dropdown
---> ERROR

## 😯 Current behavior

Error is thrown

## 🤔 Expected behavior

No error

## 🖼 Screenshots

![edit_after_edit](https://user-images.githubusercontent.com/22347393/94583504-d012e980-027d-11eb-9776-c03bb187365d.gif)

Issue #924, opened 2020-09-21
[BUG] Admin ui edit mode issue


huszerlbence:
# 🐜 Bug Report

Admin ui edit mode issue

## 📦 Package / Application

admin.test.sensenet.com AND admin.sensenet.com

## 🐜 Describe the bug

When I open a content for edit then try to open another one for edit (w/o quitting the edit mode of the first content) only the title gets changed, all other fields stay (from the first content)

check video in the screenshots section

## 👉 Steps To Reproduce

1. Go to admin.sensenet.com and log in to a repo
2. Right click on one content and select edit
3. Then w/o quitting the edit mode of the first content right click on another content in the tree and select edit

## 🤔 Expected behavior

When I edit another content the whole view should react and display the fields of the selected content.

## 🖼 Screenshots

see screen capture here:
https://sensenet-my.sharepoint.com/:v:/p/huszerlb/EUo65sEQJbJAqZKCZ3tVtL4Ba1Its2rbt6AIwWeC3_QwTQ?e=RLB5HV

PR #1012, opened 2020-10-09
Fix edit after edit error

pusztaienike:
This PR will solve both:

https://github.com/sensenet/sn-client/issues/924
https://github.com/sensenet/sn-client/issues/978

### DIFF
--- a/apps/sensenet/src/components/content/Explore.tsx
+++ b/apps/sensenet/src/components/content/Explore.tsx
@@ -99,10 +99,11 @@ export function Explore({
   const renderContent = () => {
     switch (activeAction) {
       case 'browse':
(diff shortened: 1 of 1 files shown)


apps/sensenet/src/components/content/Explore.tsx
```diff
@@ -112,6 +113,7 @@ export function Explore({
         if (contentTypeName) {
           return (
             <NewView
+              key={activeContent}
```
taki9, line 116: activeContent is not given in this case, and can not switch from one new view to another, so it can be omitted here

herflis (approved):

taki9 (approved):

PR #1013, opened 2020-10-09
Fix unit test warnings and errors

taki9:


pusztaienike (approved):

pusztaienike:
It seems that the 4 remaining warnings are from react-quill wich will be solved after update to 2.0
https://github.com/zenoamaro/react-quill/issues/509 but as of now ther is only beta version

herflis (approved):

PR #1020, opened 2020-10-13
Add E2E test for content types

shinnlok:
Few notes:

- The second test case is disabled because the feature doesn't appear to be functional
- I couldn't think of a better way to make the virtual table load everything in the list with Cypress
- Not sure where/if I should be adding `data-test` attributes

### DIFF
--- a/apps/sensenet/cypress/integration/content-types/content-types.spec.ts
+++ b/apps/sensenet/cypress/integration/content-types/content-types.spec.ts
@@ -1,0 +1,40 @@
+import { pathWithQueryParams } from '../../../src/services/query-string-builder'
+
+describe('Content types', () => {
+  beforeEach(() => {
+    cy.login()
+    cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
+      .get('div[aria-label="Content Types"]')
+      .click()
+    cy.get('.ReactVirtualized__Grid__innerScrollContainer')
+    cy.get('.ReactVirtualized__Table__Grid').scrollTo('bottom')
+    cy.xpath('//div[text()="Article"]').scrollIntoView({ duration: 500 }).as('articleRow')
+  })
+
+  it('clicking on the content types menu item should show article', () => {
+    cy.get('@articleRow').should('be.visible')
+  })
+
+  it.skip('right clicking on article should open context menu', () => {
+    cy.get('@articleRow').rightclick()
+
+    const actions = ['Edit', 'Delete', 'Set permissions']
+
+    cy.get('.MuiList-root li[role="menuitem"]')
+      .children()
+      .not('.MuiListItemIcon-root')
+      .should('have.length', actions.length)
+      .each((child, i) => {
+        expect(child.text()).to.eq(actions[i])
+      })
+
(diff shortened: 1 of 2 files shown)


apps/sensenet/cypress/integration/content-types/content-types.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { pathWithQueryParams } from '../../../src/services/query-string-builder'
+
+describe('Content types', () => {
+  beforeEach(() => {
+    cy.login()
+    cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
+      .get('div[aria-label="Content Types"]')
+      .click()
+    cy.get('.ReactVirtualized__Grid__innerScrollContainer')
```
herflis, line 9: This row seems to be not used later so you can remove it
```suggestion
```

apps/sensenet/cypress/integration/content-types/content-types.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { pathWithQueryParams } from '../../../src/services/query-string-builder'
+
+describe('Content types', () => {
+  beforeEach(() => {
+    cy.login()
+    cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
+      .get('div[aria-label="Content Types"]')
```
herflis, line 7: This selector should be changed to a `data-test=Content Types`. This attribute is already added to the drawer component by another contributor so you have to change only the selector here.

apps/sensenet/cypress/integration/content-types/content-types.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+      })
+
+    cy.get('.MuiPopover-root').click()
+  })
+
+  it('double clicking on article should open binary editor', () => {
+    cy.get('@articleRow').dblclick()
+    cy.get('div').contains('Article').should('be.visible')
+    cy.get('.monaco-editor').should('be.visible')
+    cy.get('button[aria-label="Cancel"]').click()
```
herflis, line 38: Put data-test="cancel" to the cancel button in edit-view.tsx. It is in a separate package in this repository (sn-client), you can find it in /packages/sn-controls-react/viewcontrols folder.
And then of course please change here the selector in the test for data-test

herflis (changes requested):
Yes you're right the second case cannot reproduce now, so I will remove it from the issue, thanks for pointing this out,
I've added some suggestions to your code, one is an unused row, others are related to the data-test attribute that you've mentioned. After changing the selector you will not be able to run these tests successfully on your test repository, but don't worry about it because otherwise your pr is ok 👍

apps/sensenet/cypress/integration/content-types/content-types.spec.ts
shinnlok: Done

herflis:
@shinnlok thank you for taking the time to help us! We greatly appreciate your contribution, you will get the extra sensenet-Hacktoberfest swag 🧦. We will contact you via email after the Hacktoberfest event. Hope you are having a great Hacktoberfest! 🎉 🙂

PR #1027, opened 2020-10-20
DocViewer refact

pusztaienike:
Good test file for review:
[e000.docx](https://github.com/SenseNet/sn-client/files/5511763/e000.docx)

### CHANGES

reduce requests:
- api-settings: change useEffect to useMemo --> save one request and its effect
- document-data: remove state from dep. array cause we set documentData and we check it also in useEffect
- current-children: on PreviewAvailable request we don't need "requestReload"

other:
- Thumbnails and PageList divide
- change request from "PreviewAvailable?version=" to "PreviewAvailable" if there is no version
- change "magic numbers" to 5000 (polling interval, sleeping)
- rewrite Shape.tsx from class/abstract class
- remove unneccesary props in every components
- remove unneccesary useEffect and remount the component instead
- remove SearchBar widget cause it does not work (and we don't need it)
- remove ZoomMode widget
- move rotation state from documentData to viewerState (we don't need to save rotation value)
- change activePages to activePage in viewerState
- only one comment marker (on one page) can be displayed as draft
- customZoomLevel, zoomMode removed from viewerState

### ACTION PLAN FOR THE FUTURE:
- Style should be refactored as well
eg.: style should be changeable from the top of the component
eg.: Save widget should be different in disabled state
eg.: annotations are really "ugly" now

- new feature should be implemented: 
pl. redactions, annotations, highlights, 
-> we can display them but cannot add new
(Note: IF we save a shape the rotation should be 0! - cause we don't save the image rotation form now)

One example for adding shapes:
```
const addshape = () => {
    documentData.shapes.redactions.push({ h: 100, w: 300, x: 10, y: 10, imageIndex: 1, guid: '1012' })
    documentData.shapes.highlights.push({ h: 100, w: 300, x: 10, y: 120, imageIndex: 1, guid: '1013' })
    documentData.shapes.annotations.push({
      h: 100,
      w: 300,
      x: 10,
      y: 240,
      imageIndex: 1,
      guid: '1014',
      index: 1,
      lineHeight: 17,
      text: 'balbal',
      fontBold: 'none',
      fontColor: 'black',
      fontFamily: 'Ariel',
      fontItalic: 'none',
      fontSize: '12px',
    })

    updateDocumentData(documentData)
    viewerState.updateState({ hasChanges: true })
  }
```

### UNIT TESTS

2020-12-04_14h53_35
2020-12-04_14h53_51
2020-12-04_14h54_05
2020-12-04_14h54_25


herflis (approved):

taki9 (changes requested):

packages/sn-document-viewer-react/src/hooks/usePreviewImage.ts
```diff
@@ -29,7 +23,16 @@ export const usePreviewImage = (pageNo: number) => {
           version: viewerSettings.version,
         })
         if (previewImageData?.PreviewAvailable) {
-          setPreviewImage(previewImageData)
+          if (previewImageData.PreviewImageUrl) {
+            const newImages = [...images.imageData]
+            const oldValueIndex = newImages.findIndex((image) => image.Index === pageNo)
```
taki9, line 28: do it with images.imageData.map

packages/sn-document-viewer-react/src/hooks/usePreviewImage.ts
```diff
@@ -41,14 +44,17 @@ export const usePreviewImage = (pageNo: number) => {
       }
     }
 
-    if (previewImage && !previewImage.PreviewImageUrl) {
+    if (
+      images.imageData.find((i) => i.Index === pageNo) &&
+      !images.imageData.find((i) => i.Index === pageNo)?.PreviewImageUrl
```
taki9, line 49: store in a variable, should not iterate twice for the data

taki9 (changes requested):
- after opening a doc with multiple pages, isPreviewAvailable requests are out of control on scoll :D

- uploaded a test document, opened it, clicked on generate button
  - after this I waited a minute or two for pageCount change from -1 but it not happened, so closed the doc and reopened it: 
empty-page
data request fired ca. per second, but empty page is shown

packages/sn-client-core/src/Repository/Preview.ts
```diff
@@ -22,14 +22,17 @@ export class Preview {
...
     page: number
     abortController?: AbortController
   }) {
     const responseBody = await this.repository.executeAction<{ page: number }, PreviewImageData>({
       idOrPath: options.document.idOrPath,
       method: 'POST',
-      name: `PreviewAvailable?version=${options.version}`,
+      name: 'PreviewAvailable',
+      oDataOptions: {
+        version: options.version && options.version,
```
taki9, line 34:
```suggestion
        version: options.version
```

taki9 (changes requested):

packages/sn-document-viewer-react/src/hooks/usePreviewImage.ts
pusztaienike: cannot do with map :(

taki9 (changes requested):
- comment marker should point to the same part of the document after rotation
- image only load after scrolling even if PreviewAvailable request had response before it
- open a large document with a lot of pages, go to one of the last pages with the page input: a lot of PreviewAvailable request fired, we should only request for the given page (and maybe 1 or 2 around).
- a little debouce would be great for PreviewAvailable on scrolling (try to scroll with the browser scrollbar on a large document, and a lot of request fired)

taki9 (approved):

PR #1047, opened 2020-10-26
Dependency upgrade

herflis:
- date-io is skipped as usual
- react and react-dom cannot be updated to 17.x yet because all of the tests are failing after the update (guess it is related to @types/react and @types/react-dom and that they don't have a 17.x version yet)
- ~cheap-module-eval-source-map is changed to eval-cheap-module-source-map [https://webpack.js.org/configuration/devtool/](https://webpack.js.org/configuration/devtool/)~
- ~webpack-dev-server is removed, because the new version of webpack (v5) supports serve out of the box [https://github.com/webpack/webpack-dev-server/issues/2759](https://github.com/webpack/webpack-dev-server/issues/2759)~ 
 We should wait for [this](https://github.com/jantimon/html-webpack-plugin/issues/1527) before upgrading webpack to v5


pusztaienike (approved):
🐶

taki9 (approved):

Issue #917, opened 2020-09-16
🧪 [E2E test] Add Cypress test - Upload


pusztaienike:
**_Before you start working on an issue please take a look at our [dedicated hacktoberfest site](https://hacktoberfest.sensenet.com/#eventDetails) to learn the basic rules and find additional resources to help you get started._**

# 🧪E2E test cases

The scope of these tests is to ensure that content upload works as it is intended.

![image](https://user-images.githubusercontent.com/22347393/93335010-1865ec80-f826-11ea-8928-d330ca1353ea.png)

## 😎 Role

All test should run as admin.

# Test case 1

## 🧫 Purpose of the test

To ensure uploading a content works properly.

## 🐾 Steps

1. Login with admin role
2. Click on Content menu item
3. Click on IT Workspace
4. Click on Document Library (under IT Workspace)
5. Click on 'Add new' button
6. Select Upload from the list
7. Click on the highlighted area of the dialog and select a file (OR drag & drop your file to the highlighted area.)
Test file: 
[a000.docx](https://github.com/SenseNet/sn-client/files/5232028/a000.docx)

![image](https://user-images.githubusercontent.com/22347393/93335340-95916180-f826-11ea-9297-5480c6069adf.png)

8. **Step:** Click upload
    **Expected result:** Upload completed text should appear next to the file.

![image](https://user-images.githubusercontent.com/22347393/93335484-cf626800-f826-11ea-9a8f-e19eb372bf0b.png)

9. **Step:** Close the dialog  with the ✖️ button on the top-right corner.
**Expected result:** a000.docx file should be visible on the grid next to the Chicago and Munich folders.

![image](https://user-images.githubusercontent.com/22347393/93335687-223c1f80-f827-11ea-8665-5a997799eef6.png)

10. Right click on  a000.docx 
11. Select delete from the list 
12. **Step:** Tick permanently and click on Delete button
     **Expected result:** a000.docx file should be not visible on the grid

NileshPatel17:
@herflis please assign this to me.

PR #1051, opened 2020-10-31
e2e test for upload

NileshPatel17:
@pusztaienike Added e2e test case for #917  .Please review and let me know if any changes are required.

### DIFF
--- a/apps/sensenet/cypress/integration/upload/upload.spec.ts
+++ b/apps/sensenet/cypress/integration/upload/upload.spec.ts
@@ -1,0 +1,38 @@
+import { pathWithQueryParams } from '../../../src/services/query-string-builder'
+
+describe('Upload', () => {
+  before(() => {
+    cy.login()
+    cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
+  })
+
+  it('uploading a file/content should works properly', () => {
+    const fileToBeUploaded = 'data.json'
+    cy.get('[data-test="Content"]').click()
+    cy.get('[data-test="IT Workspace"]').click()
+    cy.get('[data-test="Document Library"]').click()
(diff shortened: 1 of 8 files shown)


apps/sensenet/cypress/integration/upload/upload.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  before(() => {
+    cy.login()
+    cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
+  })
+
+  it('uploading a file/content should works properly', () => {
+    const fileToBeUploaded = 'data.json'
+    cy.get('[data-test="Content"]').click()
+    cy.get('[data-test="IT Workspace"]').click()
+    cy.get('[data-test="Document Library"]').click()
```
pusztaienike, line 13:
```suggestion
    cy.get('[data-test="Document library"]').click({ force: true })
```
Document library has 'l' instead of 'L' (not uppercased). And xou should use force click here, cause the element has CSS pointer-events: none.

apps/sensenet/src/components/AddButton.tsx
```diff
@@ -254,7 +254,7 @@ export const AddButton: React.FunctionComponent<AddButtonProps> = (props) => {
                       queryParams: { 'content-type': childType.ContentTypeName },
                     })
                   }}
-                  data-test="listitem">
+                  data-test={`${childType.DisplayName}`}>
```
pusztaienike, line 257: You can change the data-test attribute of an element, if no other test using it. In this case the attribute is already used in AddNewMenu.spec.ts (twice: line30 and line45). Please update this test also. Thanks!

NileshPatel17:
@pusztaienike Changes are done as requested.

herflis:
Thank you @NileshPatel17

Issue #796, opened 2020-08-19
🧪 [E2E test] Add Cypress test - Breadcrumb


herflis:
**_Before you start working on an issue please take a look at our [dedicated hacktoberfest site](https://hacktoberfest.sensenet.com/#eventDetails) to learn the basic rules and find additional resources to help you get started._**

# 🧪E2E test cases

The scope of these tests is to ensure that the breadcrumb displayed and works as it is intended.

## 😎 Role

All test should run as admin.

# Test case 1

## 🧫 Purpose of the test

To ensure that breadcrumb after navigating to IT doclib is displayed as it should.

## 🐾 Steps

1. Login with admin role
2. Click on _Content_ menu item
3. Open _IT workspace_ with clicking on it in the tree
**Expected result**: children of _IT Workspace_ are listed (_Calendar_ is the first child item in the grid).
4. Choose _Document library_ from _IT Workspace_'s children and open it
**Expected result**: the breadcrumb contains the following items: Content, IT Worksapce, Document library

![image](https://user-images.githubusercontent.com/6968860/90602697-763cef80-e1fa-11ea-8743-588bc1c5b2f5.png)

# Test case 2

## 🧫 Purpose of the test

To ensure that clicking on parent item in the breadcrumb opens the chosen container.

## 🐾 Steps

1. Login with admin role
2. Click on _Content_ menu item
3. Open _IT workspace_ with clicking on it in the tree
**Expected result**: children of _IT Workspace_ are listed. (_Calendar_ is the first child item in the grid).
4. Choose _Document library_ from _IT Workspace_'s children and open it
5. Click on _IT Workspace_ in the breadcrumb
**Expected result**: children of _IT Workspace_ are listed (_Calendar_ is the first child item in the grid).

# Test case 3

## 🧫 Purpose of the test

To ensure that right click on a breadcrumb item opens its actionmenu.

## 🐾 Steps

1. Login with admin role
2. Click on _Content_ menu item
3. Open _IT workspace_ with clicking on it in the tree
**Expected result**: children of _IT Workspace_ are listed. (_Calendar_ is the first child item in the grid).
4. Choose _Document library_ from _IT Workspace_'s children and open it
5. Right-click on _IT Workspace_ in the breadcrumb
**Expected result**:  Actionmenu is open for _IT Workspace_

![image](https://user-images.githubusercontent.com/6968860/90603253-4e01c080-e1fb-11ea-85b0-20ef63a6559b.png)

[...]

PR #1070, opened 2020-11-02
added e2e test case for breadcrumb

NileshPatel17:
issue #796

### DIFF
--- a/apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
+++ b/apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
@@ -6,9 +6,16 @@ describe('Breadcrumb', () => {
     cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
   })
   it('breadcrumb after navigating to IT Workspace, breadcrumb should be displayed as it should.', () => {
-    const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace']
+    const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace', '/', 'Document Library']
     cy.get('[data-test="drawer-menu-item-Content"]').click()
-    cy.get('[data-test="IT Workspace"]').click()
+    cy.get('[data-test="IT Workspace"]')
+      .click()
+      .then(() => {
+        cy.get('[data-test="Calendar"]').should('be.visible')
+        cy.get('[data-test="table-cell-Calendar"]').should('be.visible')
+      })
+
+    cy.get('[data-test="Document Library"]').click({ force: true })
     cy.get('nav[aria-label="breadcrumb"] li').each(($el) => {
       expect(expectedBreadcrumbItems).to.include($el.text())
     })
@@ -17,15 +24,17 @@ describe('Breadcrumb', () => {
     const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace', '/', 'Document Library']
     cy.get('[data-test="drawer-menu-item-Content"]').click()
     cy.get('[data-test="IT Workspace"]').click()
-    cy.get('[data-test="Document Library"]').click()
+    cy.get('[data-test="Document Library"]').click({ force: true })
     cy.get('nav[aria-label="breadcrumb"] li').each(($el) => {
       expect(expectedBreadcrumbItems).to.include($el.text())
     })
+    cy.get('[data-test="Document Library"]').should('be.visible')
+    cy.get('[data-test="table-cell-Document Library"]').should('be.visible')
   })
   it('right click on a breadcrumb item should open its actionmenu.', () => {
     cy.get('[data-test="drawer-menu-item-Content"]').click()
     cy.get('[data-test="IT Workspace"]').click()
-    cy.get('[data-test="Document Library"]').click()
+    cy.get('[data-test="Document Library"]').click({ force: true })
     cy.get('nav[aria-label="breadcrumb"] li')
       .find('button[aria-label="IT Workspace"]')
       .rightclick()
--- a/apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
+++ b/apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
@@ -1,41 +1,56 @@
 import { pathWithQueryParams } from '../../../src/services/query-string-builder'
 
+const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace', '/', 'Document library']
+
 describe('Breadcrumb', () => {
   before(() => {
     cy.login()
     cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
   })
   it('breadcrumb after navigating to IT Workspace, breadcrumb should be displayed as it should.', () => {
-    const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace', '/', 'Document Library']
     cy.get('[data-test="drawer-menu-item-Content"]').click()
     cy.get('[data-test="IT Workspace"]')
       .click()
       .then(() => {
-        cy.get('[data-test="Calendar"]').should('be.visible')
         cy.get('[data-test="table-cell-Calendar"]').should('be.visible')
       })
 
-    cy.get('[data-test="Document Library"]').click({ force: true })
-    cy.get('nav[aria-label="breadcrumb"] li').each(($el) => {
-      expect(expectedBreadcrumbItems).to.include($el.text())
-    })
+    cy.get('[data-test="Document library"]').click({ force: true })
+    cy.get('nav[aria-label="breadcrumb"] li')
+      .should('have.length', expectedBreadcrumbItems.length)
+      .each(($el) => {
+        expect(expectedBreadcrumbItems).to.include($el.text())
+      })
   })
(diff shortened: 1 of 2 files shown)


apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    cy.login()
+    cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
+  })
+  it('breadcrumb after navigating to IT Workspace, breadcrumb should be displayed as it should.', () => {
+    const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace']
+    cy.get('[data-test="drawer-menu-item-Content"]').click()
+    cy.get('[data-test="IT Workspace"]').click()
+    cy.get('nav[aria-label="breadcrumb"] li').each(($el) => {
+      expect(expectedBreadcrumbItems).to.include($el.text())
+    })
```
pusztaienike, line 14: Missing steps:

- Open IT workspace with clicking on it in the tree
- Expected result: children of IT Workspace are listed (Calendar is the first child item in the grid).
- Choose Document library from IT Workspace's children and open it
- Expected result: the breadcrumb contains the following items: Content, IT Worksapce, Document library

apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
+  it('clicking on parent item in the breadcrumb should open the chosen container', () => {
+    const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace', '/', 'Document Library']
+    cy.get('[data-test="drawer-menu-item-Content"]').click()
+    cy.get('[data-test="IT Workspace"]').click()
```
pusztaienike, line 19: Missing step:
Expected result: children of IT Workspace are listed. (Calendar is the first child item in the grid).

apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    cy.get('[data-test="Document Library"]').click()
+    cy.get('nav[aria-label="breadcrumb"] li').each(($el) => {
+      expect(expectedBreadcrumbItems).to.include($el.text())
+    })
```
pusztaienike, line 23: This check is not needed here. But you missed some steps:

- Click on IT Workspace in the breadcrumb
- Expected result: children of IT Workspace are listed (Calendar is the first child item in the grid).

apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
+  it('right click on a breadcrumb item should open its actionmenu.', () => {
+    cy.get('[data-test="drawer-menu-item-Content"]').click()
+    cy.get('[data-test="IT Workspace"]').click()
```
pusztaienike, line 27: Missing step:
Expected result: children of IT Workspace are listed. (Calendar is the first child item in the grid).

apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    cy.get('[data-test="Document Library"]').click()
```
pusztaienike, line 28:
```suggestion
    cy.get('[data-test="Document Library"]').click({ force: true })
```

apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    cy.get('[data-test="IT Workspace"]').click()
+    cy.get('nav[aria-label="breadcrumb"] li').each(($el) => {
+      expect(expectedBreadcrumbItems).to.include($el.text())
+    })
+  })
+  it('clicking on parent item in the breadcrumb should open the chosen container', () => {
+    const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace', '/', 'Document Library']
+    cy.get('[data-test="drawer-menu-item-Content"]').click()
+    cy.get('[data-test="IT Workspace"]').click()
+    cy.get('[data-test="Document Library"]').click()
```
pusztaienike, line 20:
```suggestion
    cy.get('[data-test="Document Library"]').click({ force: true })
```

pusztaienike (commented):
The test does not follow the test case. There is no check for the grid, and there are a lot of missing steps. I feel some misunderstanding here.  Let me please illustrate the page with a picture:
image
You missed to check the grid in some cases. I hope that this pic can help a little.

Thanks :)

apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
NileshPatel17: should not "Expected result: children of IT Workspace are listed (Calendar is the first child item in the **grid**)." be "Expected result: children of IT Workspace are listed (Calendar is the first child item in the **list**)?? I can see Calendar in the list, not in the grid. please correct me if  i am wrong.

pusztaienike: @NileshPatel17,
You have to see them in both places (in the grid and in the list as well). But it is enough to check them in the grid regarding to the test case. See the attached gif.
nileshp

apps/sensenet/cypress/integration/breadcrumb/breadcrumb.spec.ts
NileshPatel17: this is step 4:
**Expected result**: the breadcrumb contains the following items: Content, IT Worksapce, Document library

NileshPatel17:
@pusztaienike changes done.

pusztaienike:
> @pusztaienike changes done.

I have still found some difference. Here is my final version which is working good:

```
import { pathWithQueryParams } from '../../../src/services/query-string-builder'

const expectedBreadcrumbItems = ['Content', '/', 'IT Workspace', '/', 'Document library']

describe('Breadcrumb', () => {
  before(() => {
    cy.login()
    cy.visit(pathWithQueryParams({ path: '/', newParams: { repoUrl: Cypress.env('repoUrl') } }))
  })
  it('breadcrumb after navigating to IT Workspace, breadcrumb should be displayed as it should.', () => {
    cy.get('[data-test="drawer-menu-item-Content"]').click()
    cy.get('[data-test="IT Workspace"]')
      .click()
      .then(() => {
        cy.get('[data-test="table-cell-Calendar"]').should('be.visible')
      })

    cy.get('[data-test="Document library"]').click({ force: true })
    cy.get('nav[aria-label="breadcrumb"] li')
      .should('have.length', expectedBreadcrumbItems.length)
      .each(($el) => {
        expect(expectedBreadcrumbItems).to.include($el.text())
      })
  })
  it('clicking on parent item in the breadcrumb should open the chosen container', () => {
    cy.get('[data-test="drawer-menu-item-Content"]').click()
    cy.get('[data-test="IT Workspace"]')
      .click()
      .then(() => {
        cy.get('[data-test="table-cell-Calendar"]').should('be.visible')
      })
    cy.get('[data-test="Document library"]')
      .click({ force: true })
      .then(() => {
        cy.get('[data-test="table-cell-Calendar"]').should('not.be.visible')
      })
    cy.get('nav[aria-label="breadcrumb"] li')
      .find('button[aria-label="IT Workspace"]')
      .click()
      .then(() => {
        cy.get('[data-test="table-cell-Calendar"]').should('be.visible')
      })
  })
  it('right click on a breadcrumb item should open its actionmenu.', () => {
    cy.get('[data-test="drawer-menu-item-Content"]').click()
    cy.get('[data-test="IT Workspace"]')
      .click()
      .then(() => {
        cy.get('[data-test="table-cell-Calendar"]').should('be.visible')
      })
    cy.get('[data-test="Document library"]').click({ force: true })
    cy.get('nav[aria-label="breadcrumb"] li')
      .should('have.length', expectedBreadcrumbItems.length)
      .find('button[aria-label="IT Workspace"]')
      .rightclick()
      .then(() => {
        cy.get('ul[role="menu"]').should('be.visible')
      })
  })
})
```
Global:
- I have moved the expectedBreadcrumbItems to upper scope cause I used it from more tests
- *Document Library* has been changed to *Document library*
Test1:
- `.should('have.length', expectedBreadcrumbItems.length)` --> After clicking on Document library the breadcrumb renders more time: the first time it only has ''Content', '/', 'IT Workspace'' but the test won't fail cause the expectedBreadcrumbItems includes these values. You need to check the length as well. In cypress if you have an assertion the test will wait till your condition will be true, so it is a valid solution to wait for rerenders as well. :)
Test2:
- here we don't need to check the breadcrumb, just the grid itself. 
- I also added a plus step here: 
```
    cy.get('[data-test="Document library"]')
      .click({ force: true })
      .then(() => {
        cy.get('[data-test="table-cell-Calendar"]').should('not.be.visible')
      })
```
--> Here I would like to check the clicking on document library causes a rerender of the grid.
Test3:
- assertion added to wait for breadcrumb
- check of the grid has been added

Could you please update your test with my suggestions? 
Thank you for your work! 🥇

PR #1071, opened 2020-11-03
Field controls rework

taki9:


pusztaienike:
1. It seems that if a quill is empty there is no message appears like "No value set". Is it possible to add this?

image
image

2. The checkbox has the default primary color. Could it be #26a69a?
image

pusztaienike (approved):
Can we add padding here to the left-side?

2020-11-10_10h31_05

2020-11-10_10h32_04

herflis:
Checkboxes cannot be unchecked
checkbox

pusztaienike (approved):

taki9:
> Checkboxes cannot be unchecked
> checkbox

fixed

herflis:
something is not okay with the field control docs on the storybook page

image

Issue #1040, opened 2020-10-23
[BUG] Rendering errors on SETUP page


kavics:
# 🐜 Bug Report

There are more rendering errors on the Setup page.
1 - Clipped titles: "DocumentPreview.settin",  "TaskManagement.settin"
2 - Invisible and multiline buttons.

## 👉 Steps To Reproduce

1 - Log in to https://admin.test.sensenet.com repo: https://snover.service.sensenet.com.
2 - In the left menu click Setup.
3 - See the boxes (screenshot)

## 🤔 Expected behavior

1 - The titles should not be cropped.
2 - The buttons should be always visible.
3 - The space among the buttons should be proportional.
4 - A button should always be a single line.

## 🖼 Screenshots

![5-settings](https://user-images.githubusercontent.com/2006057/97060153-c76ab600-1592-11eb-9807-c509f10dbade.png)

## 🌍 Your environment (for apps / examples)

- Browser:
  Chrome
- Version:
  86.0.4240.75

herflis:
@kavics I cannot reproduce this issue, can you please confirm that the issue is gone?

![image](https://user-images.githubusercontent.com/6968860/97271344-b9d25c00-1830-11eb-933e-8ffc474dec82.png)

taki9:
On setup page height of boxes in a row should always be equal.
(setup.tsx line 95 should align items as stretch,

Issue #1039, opened 2020-10-23
[BUG] TreeView level overlay


kavics:
# 🐜 Bug Report

When I open a level in the TreeView the main level's next items slip on the opened child level. The screenshot shows the bug.

## 👉 Steps To Reproduce

1 - Log in to https://admin.test.sensenet.com repo: https://snover.service.sensenet.com.
2 - In the tree click (open) the explorer (last item) / Content / Smartfolders
3 - Scroll down

## 🖼 Screenshots

![3-treeview-linespaces](https://user-images.githubusercontent.com/2006057/97058564-75279600-158e-11eb-8cb6-eb4af89622c6.png)

## 🌍 Your environment (for apps / examples)

- Browser:
Chrome 1392x984
- Version:
86.0.4240.75

herflis:
@kavics I cannot reproduce this issue after the upgrade, can you please confirm, that the issue is gone?
![image](https://user-images.githubusercontent.com/6968860/97270872-1da85500-1830-11eb-8dd1-9a473d0dcbc0.png)

PR #1075, opened 2020-11-06
Fix tree and learnmore on chrome fontsize change

pusztaienike:
#1040 
#1039


herflis (approved):

taki9:
On setup page height of boxes in a row should be equal.
(setup.tsx line 95 should align items as stretch)

taki9 (commented):
On setup page height of boxes in a row should be always equal.
(setup.tsx line 95 should align items as stretch)

apps/sensenet/src/components/setup/setup.tsx
```diff
@@ -92,7 +92,7 @@ const Setup = () => {
         return (
           <>
             {settings.length ? (
-              <div className={globalClasses.centered} style={{ flexWrap: 'wrap' }}>
+              <div className={globalClasses.centeredHorizontal} style={{ alignItems: 'strech', flexWrap: 'wrap' }}>
```
taki9, line 95: stretch, but it can be removed, because it is the default value of align-items

taki9 (approved):

PR #1106, opened 2020-12-15
Dependency upgrade

taki9:
- date-io is skipped as usual (material-ui date picker only work with the currently installed version)
- react and react-dom cannot be updated to 17.x yet, we are waiting for http://github.com/enzymejs/enzyme/issues/2429 this issue. related PR: http://github.com/enzymejs/enzyme/pull/2430
- webpack cannot be updated, related issue: https://github.com/jantimon/html-webpack-plugin/issues/1527
- new versions of copy-webpack-plugin and terser-webpack plugin are only compatible with webpack 5 (update is blocked by webpack update)

webpack cli currently not working with --open argument: https://github.com/webpack/webpack-cli/issues/2001
I moved this option to the webpack config file, works fine now, but maybe we should revert it after this issue is resolved.


herflis (approved):

pusztaienike (approved):
🎄

PR #1126, opened 2021-01-06
Picker

taki9:


pusztaienike (commented):
Nice! 🤩 

**Some notes:**

style:
- tree-view icon style is ugly :D
- checkbox should have primary color
- cancel button should have border

logic:
- every item has the same icon they should be different depend on content type
- if I type something to search input then remove it the page is empty 
(I guess that the treeview should be active here) ?? 
- picker has a header with the text 'Pick a content'
(It is now 'Reference picker')

pusztaienike (approved):
🥇

herflis (approved):

PR #1143, opened 2021-01-19
e2e tests for trash

taki9:


pusztaienike (changes requested):
The last step should prove that "test" item does not exist in the trash. 
But it checks right after clicking on the trash (when loading page appears)

Here we will need an extra step to wait for the trash page.

image

pusztaienike (approved):
🌞

herflis (approved):

PR #1161, opened 2021-01-27
Search fixes

taki9:


herflis:
- save and clear buttons are black in dark theme

buttons

- could we truncate Path for single line plus ellipsis and put the whole path into a tooltip?
- Modified By column is empty

image

herflis (approved):

pusztaienike (approved):
🐌

PR #1172, opened 2021-02-04
Webhooks page of Settings menu

pusztaienike:


apps/sensenet/src/components/field-controls/webhook-headers.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const [actualValue, setActualValue] = useState<string>()
+
+  switch (props.actionName) {
+    case 'edit':
+    case 'new':
+      return (
+        <>
+          <div>
+            <label>{localization.webhooksHeader.headers}</label>
+          </div>
```
taki9, line 43:
```
<InputLabel shrink htmlFor={`${headerKey}-key`}>
  {localization.webhooksHeader.headers}
</InputLabel>
```

apps/sensenet/src/components/field-controls/webhook-payload.tsx
```diff
@@ -0,0 +1,71 @@
...
+  switch (props.actionName) {
+    case 'edit':
+    case 'new':
+    case 'browse':
+    default:
+      return (
+        <>
+          <div>
+            <label>{localization.webhooksPayload.payload}</label>
+          </div>
```
taki9, line 42:
```
 <FormLabel style={{ transform: 'translate(0, 1.5px) scale(0.75)', transformOrigin: 'top left' }}>
  {localization.webhooksPayload.payload}
</FormLabel>
```

apps/sensenet/src/components/field-controls/webhook-payload.tsx
```diff
@@ -0,0 +1,71 @@
...
+              label={localization.webhooksPayload.customize}
+            />
+          </RadioGroup>
+          <TextareaAutosize
+            className={classes.textArea}
+            disabled
+            aria-label="minimum height"
+            rowsMin={10}
+            cols={70}
+            value={JSON.stringify(JSON.parse(value), undefined, 4)}
```
taki9, line 66: store it in a variable with useMemo (it is a pretty slow 'convertation')

apps/sensenet/src/components/field-controls/webhook-payload.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const initialState =
+    '{"nodeId":1388,"versionId":401,"version":"V0.1.D","previousVersion":null,"versionModificationDate":"2021-02-25T12:16:58.78Z","modifiedBy":1,"path":"/Root/Content/DocLib/mydoc.docx","name":"mydoc.docx","displayName":"mydoc.docx","eventName":"Create","subscriptionId":1386,"sentTime":"2021-02-25T12:16:58.9472693Z"}'
+  const [value] = useState(initialState)
+  const [useDefault, setUseDefault] = useState<boolean>(true)
+
+  useEffect(() => {
+    props.fieldOnChange?.(props.settings.Name, value)
+    //TODO: Temporary solution. This will run on change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
taki9, line 31: remove this, it will be fired on component mount

packages/sn-client-core/src/Schemas/SchemaStore.ts
```diff
@@ -116,4 +116,16 @@ export class SchemaStore {
...
+  /**
+   * Returns a list of the content types existing on schema
+   */
+  public getTypesFromSchema(): string[] {
+    const schemasCopy = [...this.schemas.getValue()]
+    const contentTypes: string[] = []
+    schemasCopy.forEach((schema) => {
+      contentTypes.push(schema.ContentTypeName)
+    })
+    return contentTypes
```
taki9, line 129: map?

apps/sensenet/src/components/field-controls/webhook-filter.tsx
```diff
@@ -0,0 +1,564 @@
...
+              className={classes.containerSelectorInput}
+              autoFocus={props.autoFocus}
+              autoComplete="off"
+              name={props.settings.Name}
+              id={props.settings.Name}
+              label={props.settings.DisplayName}
+              placeholder={props.settings.DisplayName}
+              value={value.Path}
+              fullWidth={true}
+            />
```
taki9, line 306: should be readonly

taki9 (changes requested):
- if the value of http method is not changed from default, it will not be saved on backend. Maybe default value is not set on CTD?
- I think webhook-trigger would be a better name from webhook-filter

apps/sensenet/src/components/field-controls/webhook-filter.tsx
```diff
@@ -0,0 +1,565 @@
+/**
+ * @module FieldControls
+ */
+
+import { ReactClientFieldSetting, renderIconDefault, typeicons } from '@sensenet/controls-react'
+import { ReferencePicker } from '@sensenet/controls-react/src/fieldcontrols/reference-grid/reference-picker'
```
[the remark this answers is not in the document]
pusztaienike: it was not exported :( I had to change exports in sn-controls-react package

apps/sensenet/src/components/field-controls/webhook-headers.tsx
pusztaienike: headerKey is not available here

apps/sensenet/src/components/field-controls/webhook-filter.tsx
pusztaienike: disabled? It is not the right way I think :D 
image

pusztaienike:
> * if the value of http method is not changed from default, it will not be saved on backend. Maybe default value is not set on CTD?
> * I think webhook-trigger would be a better name from webhook-filter

1. It seems okay for me: 
image

2. The field in the Webhook CTD is also webhookFilter, so it should be consistent I think. But we can discuss it. @herflis ?

herflis:
> > * if the value of http method is not changed from default, it will not be saved on backend. Maybe default value is not set on CTD?
> > * I think webhook-trigger would be a better name from webhook-filter
> 
> 1. It seems okay for me:
>    image
> 2. The field in the Webhook CTD is also webhookFilter, so it should be consistent I think. But we can discuss it. @herflis ?

1. I think you should add DefaultValue property as well, but I'm not sure.
2. Yes, I think trigger is a better name because it is also the title of the field set on the ui

herflis (approved):

taki9 (approved):

PR #1180, opened 2021-02-05
Advanced search

taki9:


pusztaienike (commented):
I could not save a search with the following data:
keyword: 'Sample',
modificationDate: 'today'

I got 'Unexpected end of JSON input' error, but response number 200 from backend, but saved queries was empty:

request payload:
```
{query: "(Name:'Sample*' OR DisplayName:'Sample*') AND (ModificationDate:>=@@Today@@)",…}
displayName: "Search results for 'Sample'"
query: "(Name:'Sample*' OR DisplayName:'Sample*') AND (ModificationDate:>=@@Today@@)"
queryType: "Public"
uiFilters: "{"term":"Sample","filters":{"type":{"name":"all","icon":{"type":{"type":{},"compare":null},"key":null,"ref":null,"props":{},"_owner":null}},"date":{"name":"modifiedToday","query":{"field":"ModificationDate","value":"@@Today@@"}},"reference":{"name":"anybody"}}}"
```

herflis (changes requested):
- the clear button should have a tooltip
- the clear button does not clear the value of the textbox
search-clear
- (after the clear button disappeared the save query button did not work either)
- cannot navigate up in the picker from lower level containers

herflis (approved):

pusztaienike (approved):

PR #1188, opened 2021-02-11
Fix minor issues

pusztaienike:


packages/sn-controls-react/src/fieldcontrols/file-size.tsx
```diff
@@ -0,0 +1,118 @@
...
+import React, { useState } from 'react'
+import { changeTemplatedValue } from '../helpers'
+import { ReactClientFieldSetting } from './client-field-setting'
+import { defaultLocalization } from './localization'
+
+/**
+ * Field control that represents a Number field. Available values will be populated from the FieldSettings.
+ */
+export const FileSizeField: React.FC<ReactClientFieldSetting<NumberFieldSetting>> = (props) => {
+  const localization = deepMerge(defaultLocalization.number, props.localization?.number)
```
taki9, line 19: each field control should have its own localization

packages/sn-controls-react/src/fieldcontrols/file-size.tsx
```diff
@@ -0,0 +1,118 @@
...
+    case 'browse':
+    default:
+      return (
+        <div>
+          <Typography variant="caption" gutterBottom={true}>
+            {props.settings.DisplayName}
+          </Typography>
+          <Typography variant="body1" gutterBottom={true}>
+            {props.fieldValue != null ? (
+              <>{props.fieldValue && returnValueWithUnit(toNumber(props.fieldValue)!)}</>
```
taki9, line 110: remove `props.fieldValue &&`
existance is checked above and we should handle 0

packages/sn-controls-react/src/fieldcontrols/file-size.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return Number.isInteger(toNumber(props.fieldValue)!) || props.settings.Type === 'IntegerFieldSetting' ? 1 : 0.1
+  }
+
+  const round = (num: number, precision = 1) => {
+    const multiplier = Math.pow(10, precision)
+    return Math.round((num + Number.EPSILON) * multiplier) / multiplier
+  }
+
+  const returnValueWithUnit = (fieldValueNumber: number) => {
+    const inKiloBytes = round(fieldValueNumber / 1024)
```
taki9, line 54: it is a little ugly solution, maybe could be better with recursion?

taki9 (commented):
create unit tests for the new control

packages/sn-controls-react/src/fieldcontrols/file-size.tsx
pusztaienike: I would write  'No value set' when fieldvalue is undefined, null or has the value of 0

pusztaienike: I would write 'No value set' when the value is 0 (cause file size cannot be exact 0 I think)

herflis (approved):

taki9 (approved):

PR #1192, opened 2021-02-16
Change moment to date-fns

pusztaienike:
image


apps/sensenet/src/hooks/use-date-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+  const personalSettings = usePersonalSettings()
+
+  const formatDistanceFromNow = (date: any) => {
+    const isHungarian = personalSettings.language === 'hungarian'
+    return formatDistanceToNow(date, { locale: isHungarian ? hu : enUS, addSuffix: true })
+  }
+
+  return {
+    /**
+     * Parse date
```
taki9, line 31: you can remove these comments here, don't give additional info

apps/sensenet/src/components/view-controls/browse-view.tsx
```diff
@@ -55,6 +58,7 @@ export const BrowseView: React.FC<BrowseViewProps> = (props) => {
         handleCancel={() => navigateToAction({ history, routeMatch })}
         controlMapper={controlMapper}
         localization={{ cancel: localization.forms.cancel }}
+        locale={personalSettings.language === 'hungarian' ? hu : enUS}
```
taki9, line 61: we should create a function on LocalizationContext, which returns the locale (this approach is not sustainable if we have more languages available).

pusztaienike: I used an object there to define the languages, the langcodes and the locale. I hope it is also good :)

apps/sensenet/src/hooks/use-date-utils.ts
```diff
@@ -0,0 +1,45 @@
+import format from 'date-fns/format'
+import formatDistanceToNow from 'date-fns/formatDistanceToNow'
+import enUS from 'date-fns/locale/en-US'
+import hu from 'date-fns/locale/hu'
+import parseISO from 'date-fns/parseISO'
+import { usePersonalSettings } from '.'
+
+export const parseDate = (date: any) => {
```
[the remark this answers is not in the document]
pusztaienike: It should be number | Date than, cause the date-fns functions expect the parameter with that types:
image

taki9 (approved):

herflis (approved):

PR #1208, opened 2021-02-19
Optimize and lazy load monaco editor

taki9:
- Which (programming) languages should be supported?
- Which features should be enabled?
```
all disabled: 1.54 MB (379.62 KB gzipped)
all enabled: 2.27 MB (560.32 KB gzipped)
```

- execute action dialog loading?


herflis:
`json `and `xml `as languages and `hover`, `find `and `suggest `as features are enough for me now

herflis (approved):

pusztaienike (approved):
🌞 🌻

PR #1211, opened 2021-02-22
Remove sn6 icons from sn-icons-react package

taki9:


herflis (approved):

pusztaienike (commented):
I don't know if it is related to this, but some cases we use different icons for the same content type: e.g.: sample ws, image library etc.
image

taki9:
> I don't know if it is related to this, but some cases we use different icons for the same content type: e.g.: sample ws, image library etc.
> image

No, it is not related. But worth a new issue.

pusztaienike:
> > I don't know if it is related to this, but some cases we use different icons for the same content type: e.g.: sample ws, image library etc.
> > image
> 
> No, it is not related. But worth a new issue.

https://github.com/SenseNet/sn-client/issues/1219 done

pusztaienike (approved):

PR #1214, opened 2021-02-24
Dependecy upgrade february

pusztaienike:
### SKIPPED PACKAGES
- date-io is skipped as usual (material-ui date picker only work with the currently installed version)
- react and react-dom cannot be updated to 17.x yet, we are waiting for http://github.com/enzymejs/enzyme/issues/2429 this issue. related PR: http://github.com/enzymejs/enzyme/pull/2430
- husky https://github.com/typicode/husky/issues/854
- typescript and eslint (and any related stuff with eslint) same here: https://github.com/nuxt/typescript/issues/489
- eslint-config-prettier (lint is not working with the new version I have checked the docs, chaged the eslint.rc as well)
https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md

### NOTE
Don't delete the yarn.lock and generate it again cause one of our dependency will use the new version of typescript
which is incompatible with eslint and prettier

**Webpack has been updated to v5.x. Please check the configs twice before we merge it to develop!**
migration guide: https://webpack.js.org/migrate/5/


herflis (approved):

taki9 (commented):
> When using [hash] placeholder in webpack configuration, consider changing it to [contenthash]. It is not the same, but proven to be more effective.

search for [hash:8] on the project and replace it with [contenthash:8]

packages/sn-search-react/src/Components/Fields/ReferenceField.tsx
```diff
@@ -177,7 +177,7 @@ export function ReferenceField<T extends GenericContent>(props: ReferenceFieldPr
           displayName={displayName}
           description={description}
           helperText={helperText}
-          inputProps={inputComponentProps}
+          inputProps={inputComponentProps as any}
```
taki9, line 180: i think `as any` is not necessary

pusztaienike: :'(
image

taki9 (approved):

PR #1225, opened 2021-03-04
Fix DMS issues

herflis:
- minor dms related fixes
- example app clientId-s are changed to the external one of the dev.demo


pusztaienike (commented):
**I found a lot of problems with DMS demo, maybe we can fix some of them till the refactor:**

1. There is an unvisible batch icon (Share)

image


2. Something is wrong with batch delete. If I create a new folder under Document library and select the newly created folder, then click on batch delete action:
It would like to delete the whole Document library (If I submit it will be deleted) 
The correct behaviour should be: Only selected items would be deleted, not their parent.

dms_demo


3. Copy has no error-handling (i wanted to copy the newly created Test folder to Munich folder twice)
Strange cause it shoud throw error like on admin-ui (Failed to copy content Test to Munich Cannot copy the content because the target folder already contains a content named 'Test'.) but no error is shown on the ui. 

4. Moving is not working well: 
After moving a newly created folder to Chicago folder nothing happens. On network tab it seems good, but on the ui there is no change:
```
request payload: {paths: [], targetPath: "/Root/Content/IT/Document_Library/Chicago"}
response: 200 {
  "d": {
    "__count": 0,
    "results": [],
    "errors": []
  }
}
```
5. There should be a padding on dropdowns (share dialog, search filter page etc.)

image

6. 'More sharing options' and 'Copy link' are not working on share dialog
Copy link contains a link like this: http://localhost:3000/#preview/MjA2NA==

7. If I have a saved query on dev.demo and I would like to open it via dms demo, the site has crashed or do nothing is on the page.

image

**Request url:** https://dev.demo.sensenet.com/odata.svc/Root/Content/IT/('Document_Library')/GetQueries?$select=Id%2CPath%2CName%2CType%2CParentId%2CActions%2COwner%2CDisplayName%2CLocked%2CCheckedOutTo%2CApprovable%2CQuery%2CIcon&scenario=ContextMenu&metadata=no&$inlinecount=allpages&$expand=Actions%2COwner%2CCheckedOutTo&$top=10000
**I think it should be:** https://dev.demo.sensenet.com/odata.svc/Root/('Content')/GetQueries?$orderby=DisplayName%20asc&$select=Id%2CPath%2CName%2CType%2CDisplayName%2CIcon%2CIsFile%2CIsFolder%2CParentId%2CVersion%2CPageCount%2CBinary%2CCreationDate%2CAvatar%2CQuery&$expand=CheckedOutTo%2CCreatedBy&metadata=no&$inlinecount=allpages&$top=10000

pusztaienike (approved):

taki9 (approved):

PR #1236, opened 2021-03-10
Change Name to DisplayName in new-view's  title

pusztaienike:


herflis (approved):

taki9 (changes requested):
store the calculated displayname with useMemo, don't want to loop over the schema on every render

taki9 (approved):

PR #1241, opened 2021-03-16
Fix e2e tests (because of the new menu structure)

pusztaienike:


taki9 (commented):
`apps\sensenet\src\components\settings\index.tsx - line 118`

Maybe drawer-submenu-item-... or something like this would be better. These are not real drawer items.

taki9 (approved):

herflis (approved):

PR #1276, opened 2021-04-07
Image gallery rework part1

taki9:


pusztaienike (approved):
The style of the header text is different on different pages (and the shadow of the header also). Otherwise it is okay 🥇 


2021-04-12_10h30_47

2021-04-12_10h30_52

herflis (approved):

PR #1311, opened 2021-05-03
Webhook template icons

herflis:


apps/sensenet/package.json
```diff
@@ -79,6 +79,8 @@
     "webpack-merge": "^5.7.3"
   },
   "dependencies": {
+    "@iconify-icons/logos": "^1.1.9",
+    "@iconify/react": "^1.1.4",
```
taki9, line 83: yarn.lock should have changed?

apps/sensenet/src/components/AddButton.tsx
```diff
@@ -51,15 +51,15 @@ const useStyles = makeStyles((theme: Theme) => {
     },
     listDropdown: {
       padding: '10px 0 10px 10px',
-      width: '245px',
+      minWidth: '245px',
       maxHeight: '548px',
+      width: 'auto',
```
taki9, line 56: auto is its default value, so it is unnecessary to set

herflis: @taki9 without setting with auto, the dropdown has a defult with and the items are cut in half

pusztaienike (commented):
For me it does not seems good: only gatsby has the right icon
image

This branch has also lint error and it seems that github cannot deploy storybook as well. 
(I think it is related to yarn.lock upgrade ?!)

pusztaienike (approved):
🌞

taki9 (approved):

PR #1316, opened 2021-05-07
Fix switcher's position on User and WebHookSubscription types' edit and browse views

herflis:


packages/sn-controls-react/src/helpers.ts
```diff
@@ -13,11 +13,12 @@ export const changeTemplatedValue = (value: string | undefined, evaluatedValue?:
...
-    (field.fieldSettings.Name === 'Avatar' && contentType.includes('User')) ||
-    (field.fieldSettings.Name === 'Enabled' && contentType.includes('User')) ||
-    (field.fieldSettings.Name === 'Enabled' && contentType === 'WebHookSubscription') ||
+    (field.fieldSettings.Name === 'Avatar' && (contentType === 'User' || parentType === 'User')) ||
+    (field.fieldSettings.Name === 'Enabled' && (contentType === 'User' || parentType === 'User')) ||
+    (field.fieldSettings.Name === 'Enabled' &&
+      (contentType === 'WebHookSubscription' || parentType === 'WebHookSubscription')) ||
     field.fieldSettings.Type === 'LongTextFieldSetting'
   )
 }
```
taki9, line 24: should get content as parameter, instead of content.Type and call repository.schemas.isContentFromType(content, 'User) which returns a boolean value
(the same for WebHookSubscription)

taki9 (changes requested):

pusztaienike (approved):
It can be prettier with the idea in @taki9's comment, otherwise it is good 🦓

taki9 (approved):

PR #1324, opened 2021-05-12
e2e test for breadcrumb

taki9:


pusztaienike (commented):
Hmm.. honestly I don't think we need this test. We already have a test for breadcrumb, which checks almost the same stuffs: (integration/breadcrumb/breadcrumb.spec.ts)

This issue should be solved: https://github.com/sensenet/sn-client/issues/796 by this PR: https://github.com/SenseNet/sn-client/pull/1070 (merged).

pusztaienike (commented):
Navigation in the tree is not tested in your version. :(  Maybe it is not a problem cause it is almost the same like cy.visit a path, but there is no check for the grid at all. 
(step 3 is fully skipped in every test:
_Open IT workspace with clicking on it in the tree
Expected result: children of IT Workspace are listed. (Calendar is the first child item in the grid)._)

pusztaienike (approved):

herflis (approved):

PR #1334, opened 2021-05-20
Webhook browse/details view

herflis:


apps/sensenet/src/components/field-controls/webhook-payload.tsx
```diff
@@ -106,35 +117,15 @@ export const WebhookPayload: React.FC<ReactClientFieldSetting<LongTextFieldSetti
...
-            className={classes.textArea}
-            disabled
-            aria-label="minimum height"
-            rowsMin={10}
-            cols={70}
-            value={value}
-          />
+          </Typography>
+          <Typography variant="body1" gutterBottom={true}>
+            {String(useDefault) ? localization.webhooksPayload.defaultPayload : localization.webhooksPayload.customize}
```
pusztaienike, line 124: useDefault is a boolean, so it should be: (otherwise it is not working for me)
```suggestion
            {useDefault ? localization.webhooksPayload.defaultPayload : localization.webhooksPayload.customize}
```

apps/sensenet/src/components/field-controls/webhook-payload.tsx
```diff
@@ -106,35 +117,15 @@ export const WebhookPayload: React.FC<ReactClientFieldSetting<LongTextFieldSetti
...
+          </Typography>
+          <Paper elevation={0} className={widgetClasses.container}>
+            {value || ''}
```
pusztaienike, line 127: 1. It seems better if you wrap this into a pre tag:
```suggestion
           <pre>{value || ''}</pre>
```
2. We can also render the example if  default payload is selected?! cause nothing is there in this case:
image

pusztaienike (changes requested):

apps/sensenet/src/components/field-controls/webhook-payload.tsx
herflis: I think we cannot display the example as a default value because it is not the default value, if a field has no value it should be empty as in every other case :)

pusztaienike (commented):
If I checked the "trigger for all events" radio button we don't know which content types are in the trigger list (on browse view). I think we should list them in this case.
image

pusztaienike (approved):

taki9 (changes requested):
merged in #1335:

apps/sensenet/src/components/field-controls/webhook-payload.tsx
classes.paper is not used

apps/sensenet/src/components/field-controls/webhook-headers.tsx
line 129: In my opinion it would be better without rowContainer class (too large margins right now)

taki9 (approved):

PR #1345, opened 2021-06-09
Basic sensenet editor

taki9:


pusztaienike:
1. I think background color should be a little bit lighter, cause it has no contrast for me on drak mode. (On light mode there is a shadow)
image
2. I have insert a link with the tex of 'asd' but it is not visible on dark mode.
image
3. Tooltips for the icons. (it is a feature request :) )

pusztaienike (approved):

herflis (approved):

PR #1365, opened 2021-06-16
Fix storybook folder structure

pusztaienike:
There was an issue with storybook folder structure:

image
It may be cause of this update: https://github.com/SenseNet/sn-client/pull/1299 

Result:  fieldcontrols is a root folder from now

image

### DIFF
--- a/examples/sn-react-component-docs/.storybook/manager.js
+++ b/examples/sn-react-component-docs/.storybook/manager.js
@@ -1,3 +1,4 @@
+import { addons } from '@storybook/addons'
 import '@storybook/addon-notes/register'
 import '@storybook/addon-knobs/register'
 import '@storybook/addon-actions/register'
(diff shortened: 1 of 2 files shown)


herflis (approved):

examples/sn-react-component-docs/.storybook/manager.js
```diff
@@ -1,7 +1,14 @@
...
 import '@storybook/addon-links/register'
 import '@storybook/addon-options/register'
 import '@storybook/addon-a11y/register'
 import '@storybook/addon-viewport/register'
+
+addons.setConfig({
+  sidebar: {
+    showRoots: false,
+  },
+})
```
taki9, line 14: is it needed?

pusztaienike: yeah, otherwise it will display FieldControls as a root (I tried it and it was not pretty), now we want folder icon ... 
here is the example: https://storybook.js.org/docs/riot/configure/sidebar-and-urls#roots

taki9 (approved):

PR #1395, opened 2022-03-25
fix broken action on content name w parenthesis

VargaJoe:


VargaJoe:
Check former getParentPath issue (949)

VargaJoe:
Helpers like `getSegments` and `isItemPath` are used by `getContentUrlByPath` and while the former one does not handle id format, the other on tries to. I suggest to remove id handling from `isItemPath` too and make the logic consistent between the two helper.

Id format only works in `/OData.svc/Content(id)` format and id can not be used anywhere in the path. If `/Root` is in the path id will not work certainly. It's a feature not a bug. Url by id is handled by `getContentUrlbyId` anyway which does not use the segments logic.

Restobit (approved):

PR #1447, opened 2023-01-17
Change to Content Explorer

hassanad94:


apps/sensenet/src/localization/default.ts
```diff
@@ -118,7 +118,7 @@ const values = {
   drawer: {
     titles: {
       Content: 'Content',
-      ContentTypes: 'Content Types',
+      ContentTypes: 'Content Explorer',
```
tusmester, line 121: Not this one, the Content Types title was OK. We meant to change the CustomContent menu item below to Content Explorer.

tusmester (changes requested):

tusmester (approved):

PR #1452, opened 2023-01-19
1755 rearrange left panel features

hassanad94:
We have discussed that Search is not the correct name of the menu item, so I have changed it to saved queries.

I've managed to do the double icons 😁😁
There is a new attribute to an Object called "contentSetting", it is a boolean value that describes the usage of the item. If it's true then it will appear at the bottom of the drawer. I could not find a better name, so I accept every suggestion.

Due to menu name changes ( "Search" -> "Saved Query" ), I had to configure the test. Maybe I will need some help with it if it fails.


tusmester:
Suggestions for the new flag: `advancedItem` ? `systemItem` ?

VargaJoe (approved):

kavics (approved):

tusmester (approved):

tusmester (approved):

PR #1453, opened 2023-01-20
Correct Date Time Format

hassanad94:
Thats the first part of the issue.

- ( Bug fix )
- I fixed the format of the control. Format: "yyyy MMMM dd"


hassanad94:
I have more things to do, but i pushed-ed it so that i can see if any error occur.

hassanad94:
Currently im ready with:
- Bug Fix
- Format display
- Min Date

hassanad94:
- If BirthDate is the minimum value than its treated like null

Notes to discuss: 
- Currently there is not way you can set datepickers to null after you set a value.
- There is a part of the code which will never execute in the current system.

hassanad94:
### Summary for this Issue
## changes
- Date display format
- Min Date Error Fixed
- null value handle

## discuss 
- changeTemplatedValue and its usecase.

kavics (approved):

tusmester (approved):

PR #1456, opened 2023-01-25
1839 browse mode page count field 

hassanad94:


VargaJoe (approved):

packages/sn-controls-react/src/fieldcontrols/localization.ts
```diff
@@ -37,6 +37,13 @@ export const defaultLocalization = {
   },
   number: {
     noValue: 'No value set',
+    pageCountValues: {
+      '-1': 'No preview provider',
+      '-2': 'Preview generation postponed',
+      '-3': 'Preview error',
+      '-4': 'Preview generation in progress',
+      '-5': 'Preview generation in progress',
```
tusmester, line 45: Pls check if these texts and values correspond to the list in the issue.

tusmester (changes requested):

packages/sn-controls-react/src/fieldcontrols/localization.ts
hassanad94: Sorry... I will correct it !

tusmester (approved):

PR #1466, opened 2023-02-09
Dependecy update

hassanad94:


hassanad94:
image


the wait on package is unused so i delete it from the package.json

and i will update the missing dependecies.

PR #1547, opened 2023-10-12
possible fix of trimSlashes error on Content creation

VargaJoe:


VargaJoe:
cannot reproduce the original bug atm

PR #1589, opened 2024-01-11
Disable orgunits and add icons to user's allroles menu.

kavics:
Todo: prevent executing default event handlers when a disabled menu item is clicked.


VargaJoe (approved):

hassanad94 (approved):

hassanad94:
A unit testekbe besegítsek ? Amúgy fullos :D

PR #1643, opened 2025-05-09
Feature/sn auth package extraimprovements

NoelKova:


VargaJoe (approved):

NoelKova:
After building to test, collect issues if any, then fix them up and after that we can merge into develop

